PR #4, opened 2016-03-03
Add a check for ImageMagick

PalanteJon:
This shows up on the status console on Civi 4.7.  In theory, this hook is supposed to fire on `System.check` on Civi 4.6 and give a notification once a day - but it looks like that code was never written.  Regardless, this works perfectly in 4.7, and maybe one of these days I'll submit a patch to 4.6LTS to fire `hook_check` on `System.check`.

### DIFF
--- a/mosaico.php
+++ b/mosaico.php
@@ -198,3 +198,14 @@ function mosaico_civicrm_pageRun(&$page){
   }
 }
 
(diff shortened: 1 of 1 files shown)


deepak-srivastava:
@PalanteJon Thanks :+1: Merged.

PR #11, opened 2016-03-16
create static directory on install

mark-rodgers:
when you install the extension, if "civicrm/persist/contribute/static" does not exist then the static directory is created.
I also cleaned up some old commented out code I left in my previous PR as well as removing $mConfig['UPLOAD_URL'] and $mConfig['UPLOAD_DIR'] as they are no longer used.


deepak-srivastava:
@mark-rodgers thanks again :+1: 

We might have to take care of trailing slash. But fine for now.

Issue #26, opened 2016-06-07
Uploading an image shows then breaks with error


rocxa:
Images show briefly and then error:
no decode delegate for this image format `/tmp/magick-ab1Slan2' @ error/constitute.c/ReadImage/566

Answer, imagemagick install is probably duff..
http://stackoverflow.com/questions/30005612/imagemagick-no-decode-delegate-for-this-image-format

Although for me imagemajick shows all the necessary delegates and phpinfo shows all the supported formats and I still get the error :-(

Great (much needed) extension to CiviMail btw :-)

deepak-srivastava:
@rocxa Not sure we can help, but we can try setup one wamp install and see how it goes.

rocxa:
I have confirmed that imagemagick is working fine via php (creating thumbnails etc) on my server but seems to break when used via Mosaico.  It is probably going to be something silly like Mosaico looking in the wrong place/or using the wrong folders for images.

deepak-srivastava:
Does runcheck=1 detects anything (make sure to use latest release)? 

If you can identify what path/url appears wrong to you, i can probably suggest where to correct. Its mostly config vars.

rocxa:
I have discovered the problem is caused by empty index.html files that are added to all directories on our server to (i think) stop directory listings?!

the Utils.php file in the Mosaico extension doesn't look specifically for image files so get caught out by these html files (and probably any other type of file) in the image directories and tries to convert them to images.

deepak-srivastava:
@rocxa thanks for your PR.

re: index.html - may be its worth adding checks for such files when runcheck=1 in the url? I remember being hit by it.

deepak-srivastava:
Closing this ticket now. Please re-open if any issues. Thanks.

Issue #25, opened 2016-06-07
Selecting a template from the first screen tries to print it


rocxa:
The links are in this format:  /civicrm/mosaico/editor?snippet=2#twjeqag

I have a feling the ?snippet=2 means print this page? as every time I select a template the printer dialog box of my web browser opens.

rocxa:
This also happens when opening a previously saved template for editing.
http://******************/civicrm/mosaico/editor?snippet=2#twjeqag

deepak-srivastava:
Yeah we are aware of it. We ended up using it as only option where there is least amount of JS and theme interference, plus we get full screen. 

Think we 'll have to come up with new snippet similar to 2 but not print.

rocxa:
Adding an if statement to CRM/common/print.tpl is a quick fix to this problem.

`{crmRegion name='page-footer' allowCmsOverride=0}
{if strpos($tplFile, 'Mosaico') == false}
  <script type="text/javascript">
    window.print();
  </script>
{/if}
{/crmRegion}`

rocxa:
I tried creating another snippet which was a copy of print.tpl (e.g. noprint.tpl) without the window.print() but the copied print.tpl seemed to load up all of Civi's css/js and interfered with Mosacio's user interface so it was easier just to patch print.tpl instead.

deepak-srivastava:
@rocxa thanks for investigating. Any ideas how to do it without touching civi core file, like doing something within the extension that would achieve the same? May be overriding JS or tpl in worst case?

rocxa:
Adding this to the Mosaico page templates (before the print.tpl is loaded)  would stop the printing action:

  window.print = function () {
    return false;
  }

mark-rodgers:
This was resolved in PR #31

deepak-srivastava:
Thanks @rocxa

PR #31, opened 2016-08-05
2 Quick fixes

rocxa:
Fix image directories that don't need trailing slash.  #26
Disable window.print() where it isn't needed. #25


mark-rodgers:
Just cloned this PR, everything seems to look fine to me.

deepak-srivastava:
@rocxa :+1: 
@mark-rodgers :+1:

PR #41, opened 2016-09-30
Tokens to use for unsubscribe and optout in mosaico template editor

Kajakaran:
New tokens available to use in mosaico template editor for unsubscribe and optout
for unsubscribe - {action.etbUnsubscribeUrl}
for opt out - {action.etbOptOutUrl}


deepak-srivastava:
Used js to solve it. Civi tokens are supported now.

PR #54, opened 2016-10-14
PR for separate permission and directory fixes

rocxa:
This PR:

- creates a new permission 'Access CiviCRM Mosaico' so that permission can be granted to create Mosaico templates but not edit system templates or send emails.

- modifies the way the directories are defined so that uploaded images end up in the same directory that CiviMail uses for its image browser.  At the moment it seems they end up in a different directory.


rocxa:
This PR also includes a check to make sure imagemagick doesn't process images it has already processed to the correct size. and uses the previous version of the image instead.

deepak-srivastava:
@rocxa :+1:

CRM/Mosaico/Page/Index.php
```diff
@@ -25,6 +25,23 @@ function run() {
           ts('PHP extension Fileinfo not loaded or enabled')
         );
       }
+      if (empty($config->extensionsURL)) {
```
rocxa: @deepak-srivastava i have removed these lines again in https://github.com/veda-consulting/uk.co.vedaconsulting.mosaico/pull/57

Issue #130, opened 2017-03-18
Lengthen the 'name' column


kenwest:
When I try to save a template with the default name I get a database error "Data too long for column 'name' at row 1".

The name column is varchar(32) and the default name is "Mosaico Template dd/mm/yyyy, hh:mm:ss". Perhaps my local settings give a longer name than others.

32 characters seems too short. The corresponding column in the civicrm_msg_template field is varchar(255)

totten:
FWIW, in 2.x, the analogous column is `varchar(64)`. But I think it's a good point that we can make it longer. Submitted fix for 2.x as part of #131.

totten:
Looks like it was closed in 2.x and 1.x.

PR #133, opened 2017-03-25
Reroll patch from #130 for 1.x branch

kenwest:
Lengthen the name column of the civicrm_mosaico_msg_template table from 32 to 255. This commit assumes the schema is changed by hand:

1. I wasn't sure whether the developers are accepting PRs against 1.x
2. I need to confirm how the upgrade_XXXX function should be named. If I numbered it 4600 then anyone upgrading later to 2.x would be OK, but perhaps there are other upgrades?

Cheers!


deepak-srivastava:
@kenwest thanks for the patch. 
1. Think same PR also needs to go to v2.x
2. It could be upgrade_4601 for v1.x, and upgrade_4703 for v2.x. 
Don't think any update statement related to this work, would cause any issues if executed twice - say for someone upgrading from beta1 to v2.x.

Issue #134, opened 2017-04-05
Mosaico z-index is overlaying select2 widget in Drupal


meladawy:
**PROBLEM** 
We have just figured out that "Test" > "Select Group" field is not working 

![test mailing](https://cloud.githubusercontent.com/assets/2423218/24707072/8324aade-1a12-11e7-850f-b6e7a52c77a2.gif)

** REASONS ** 
This commit
https://github.com/veda-consulting/uk.co.vedaconsulting.mosaico/commit/ce97093dc45d0f539d7ce2e5923fd6116d41a805

is overriding z-index of Mosaico iframe which result in increasing the z-index of jQuery dialog. Generated z-index of select2 widget is always lower than 10000. 

![z-index](https://cloud.githubusercontent.com/assets/2423218/24707159/da7a2444-1a12-11e7-8d59-be8647e6e7c2.gif)

** SOLUTION **
Is to reduce the zindex of mosaico iframe to be something lower than select2 widget zindex

totten:
1. Nice animated gif. Clever use of the URL bar. :)

2. The zIndex was bumped to fix the appearance on WordPress (#114) -- ie https://github.com/veda-consulting/uk.co.vedaconsulting.mosaico/pull/119/commits/ce97093dc45d0f539d7ce2e5923fd6116d41a805

3. I wish `crmMosaicoIframe` weren't responsible for setting the zIndex on popups -- we originally used jQuery dialog which handled it well but got a lot of pushback from Jamie/Guanhuan (about the chrome).

meladawy:
Thank you @totten, just tested ur PR and its working! (Hoolaaayy)

PR #135, opened 2017-04-08
crmMosaicoIframe - Multiple styling fixes (#134)

totten:
Specifically:
 * Don't set zIndex. This fixes the select2 drop down (#134).
 * On WordPress sites, avoid collision with navbar. Its zIndex is too high.
 * Turn off the  body-level scroll bar while iframe is open. It's confusing/unnecessary.


meladawy:
Hey man, just merged this PR with my local development branch and its working...thanksss!

after

PR #141, opened 2017-04-26
Display and style clear icon

meladawy:
Previously i have disabled clear icon for "Reply-to" field for misunderstanding the requirements. In this PR, I'm reverting this change and changing the position of clear icon

Before:
before

After:
after


meladawy:
I'll PR from branches just to make PRs more independents

PR #144, opened 2017-04-27
HW-281: Display and style clear icon

meladawy:
Previously i have disabled clear icon for "Reply-to" field for misunderstanding the requirements. In this PR, I'm reverting this change and changing the position of clear icon

Before:
before

After:
after


totten:
Agree with this - "Reply-To" should be optional.

PR #145, opened 2017-04-27
Hw 295 step highlight

meladawy:
Removing hover highlight of disabled tabs and organize mosaico.scss file

http://g.recordit.co/QHVA9GkYMo.gif


totten:
Makes sense.

PR #146, opened 2017-04-27
HW-297: Wet wizard view as default

meladawy:
Set "bootstrap-wizard" as default value while installing Mosaico extension using 

```
./bin/setup.sh -D
```

I created a new function **default_settings** that should handle any default settings commands including the following command, which set bootstrap-wizard as default

```
 cv api setting.create mosaico_layout=bootstrap-wizard
```

This **default_settings** function is executed at the end of **do_download** function


totten:
A few things:

 1. If you use buildkit and want to make a default on your system (and have it survive uninstalls/rebuilds/etc), then you can make a file `/etc/civicrm.settings.d/100-mosaico.php` and specify `$GLOBALS['civicrm_setting']['domain']['mosaico_layout']='bootstrap-wizard';`. There's a similar way to do it per-build-type (`app/config/$BUILDTYPE/civicrm.settings.d`).
 2. This particular setting has three values: `auto`, `bootstrap-single`, `bootstrap-wizard`. The default value should always be `auto`. If the intent is to change the default appearance, then patch `CRM_Mosaico_Utils::getLayoutPath()`. 
 3. Personally, I prefer single screen. I didn't go in planning to argue about that -- it's based on feedback from several user-type people. The basic reason: (a) there's less clicking around and (b) a wizard solves the non-existent problem of having too many fields. But the topic of wizards has always been controversial. If someone can propose an unbiased test to resolve the controversy, I'd like to hear it. ;)

PR #151, opened 2017-05-10
HW-297: Sets settings when installing Mosaico extension

meladawy:
Set default layout to 'bootstrap-wizard' using Mosaico settings file instead of using cv command

### DIFF
--- a/bin/setup.sh
+++ b/bin/setup.sh
@@ -81,11 +81,6 @@ function do_gencode() {
 }
 
 ##############################
-function set_default_settings() {
-  cv api setting.create mosaico_layout=bootstrap-wizard
-}
-
-##############################
 function do_download() {
   if [ ! -d "$EXTROOT/packages" ]; then
     mkdir "$EXTROOT/packages"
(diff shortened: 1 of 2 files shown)


mosaico.php
```diff
@@ -39,6 +39,7 @@ function mosaico_civicrm_install() {
  */
 function mosaico_civicrm_postInstall() {
   _mosaico_civix_civicrm_postInstall();
+  _mosaico_civicrm_install_settings();
```
omarabuhussein: it is better to keep the order like this : 

```php
_mosaico_civicrm_install_settings();

_mosaico_civix_civicrm_postInstall();
```

omarabuhussein (changes requested):

omarabuhussein (approved):

guanhuan:
Looks good now. Please merge.

totten:
IMHO, this PR was still incorrectly written. See point #2 from the previous discussion in https://github.com/veda-consulting/uk.co.vedaconsulting.mosaico/pull/146 -- the aim is for `auto` to stay pegged to mean "the default on new installations"

meladawy:
Hey @totten, Sorry I missed this one as its got merged before we complete our discussion. I'll update this PR later today. Thanks!

PR #152, opened 2017-05-11
Updated README and set default

guanhuan:
- Updated README to structure installation options better and include v2 info
 - Set default setting via the correct method

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,20 +1,33 @@
-## uk.co.vedaconsulting.mosaico
+# uk.co.vedaconsulting.mosaico
 This extension integrates Mosaico a responsive email template editor, with CiviCRM.
-- Initial Blog Post - https://civicrm.org/blogs/parvez/a-new-beginning-for-civimail
-- Beta Blog Post - https://civicrm.org/blog/deepaksrivastava/email-template-builder-civimosaico-is-now-beta
-- Initial Video - https://vimeo.com/156633077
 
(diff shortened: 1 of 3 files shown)


guanhuan:
Hi @deepak-srivastava , 

Can we please have a pre-release of 2.0-beta2 when this is merged? 

Thanks!

veda-consulting:
Sure, Deepak is back in the office on Monday and we can get it sorted for
you then.

deepak-srivastava:
@totten @guanhuan do i need to hold beta2 till https://github.com/veda-consulting/uk.co.vedaconsulting.mosaico/pull/151 is fixed?

guanhuan:
@deepak-srivastava I don't think approach discussion is worth to hold off the whole beta 2. Truth is the updated README won't allow community users to setup Mosaico easily until we release the beta 2 (see option 1). So I would suggest to release it anyway but @totten and @meladawy can continue work out the details of the approach and I will sign off before we tag official release.

deepak-srivastava:
@guanhuan thanks.
@totten I've tagged 2.0-beta2. Should be ready for publish, once build is ready with dist.civicrm.org?

deepak-srivastava:
2.0-beta2 seems to have got published - https://civicrm.org/extensions/email-template-builder.
Date is incorrect, but it could be due to incorrect date in info file.

PR #155, opened 2017-05-16
Update release date for beta 2

guanhuan:
@deepak-srivastava Thanks for the comment in https://github.com/veda-consulting/uk.co.vedaconsulting.mosaico/pull/152. This is weird since we have beta1 at 23-03-2017 but the date in the info file was still in 2016.

I have updated the date, could you retag beta2 so it can rank at the top?

Much appreciated!


deepak-srivastava:
Done re-tagging. Also deleted published version, so it can re-publish. Thx.

deepak-srivastava:
Update: Has got published, and appears on top.

Issue #160, opened 2017-05-27
Image path issues - Social media icons are not always visible


mattwire:
Having a few issues with image paths.  Running latest version on Drupal with civicrm 4.7.19.

1. Social icons don't display in preview as it is looking for a path relative to the site root (ie. https://sitename.org/templates/versafix-1/img/social_def/facebook_ok.png).
**Workaround:** Create a symlink in the site root:
`templates -> sites/default/files/civicrm/ext/uk.co.vedaconsulting.mosaico/packages/mosaico/templates`

2. Social icons don't display in mailing public URL (eg. https://sitename.org/civicrm/mailing/view?id=125) as it is looking for a path relative to civicrm (which won't work!) (ie. https://sitename.org/civicrm/mailing/templates/versafix-1/img/social_def/facebook_ok.png)

freedomlives:
Same here-- the social icons don't work in the email or on the public mailing view.

JKingsnorth:
The social icons don't display in the preview because the block definition uses relative URLs:
https://github.com/civicrm/mosaico/blob/b2b271fe6f23c3393e1196a10fa9368329e5bb46/templates/versafix-1/template-versafix-1.html#L1402

The relative URLs are rewritten to absolute URLs when the mailing is actually sent though:
https://github.com/veda-consulting/uk.co.vedaconsulting.mosaico/blob/bc9d18cadf0984b427f533378451996b93229939/CRM/Mosaico/UrlFilter.php#L34

mattwire:
See PR #182 for social icons fix

totten:
As noted in the PR, I've been [unable to reproduce the problem](https://github.com/veda-consulting/uk.co.vedaconsulting.mosaico/pull/182#issuecomment-347683815). Can anyone say if they still have the problem -- or if there's an important difference in environment/settings?

mattwire:
@totten Give me a couple of days and I'll do some testing

mattwire:
@totten I've just tested with the latest HEAD on 2.x and I'm still seeing images fail at the first hurdle:
New mailing->Edit->Add social media block->Preview->As HTML->Social icons not shown:
![civicrm-demo mjwconsult co uk_civicrm_a_](https://user-images.githubusercontent.com/2052161/33515672-77f7100c-d75e-11e7-8fb4-845031f68f87.png)

CiviCRM 4.7.27, shoreditch:master, mosaico:2.x flexmailer:master.

[...]

PR #182, opened 2017-08-23
Fix relative image urls in templates (make social icons work properly)

### LINKED ISSUE #139: Social media icons are not visible while previewing the template
**PROBLEM**
Social media icons are visible while building the template but links are getting broken when "Test" => "Preview the template"

mosaico

**Guessing**
I believe its more related to [Mosaico library ](https://github.com/voidlabs/mosaico) and the way [versafix template](https://github.com/voidlabs/mosaico/blob/master/templates/versafix-1/template-versafix-1.html#L1398) is called from Mosaico CiviCRM extension

https://github.com/voidlabs/mosaico/blob/master/templates/versafix-1/template-versafix-1.html#L1398

I believe calling template-versafix-1.html somehow from iframe instead of using modal should resolve this.

mattwire:
This should fix all image URLs that are specified as relative to the template directory (eg. img/facebook.png).  When submitted for preview/send they are rewritten to templates/versafix-1/img/facebook.png - this patch rewrites that format to an absolute format.  Should fix #139 and final part of #160.  @JKingsnorth can you verify?

### DIFF
--- a/mosaico.php
+++ b/mosaico.php
@@ -308,6 +308,24 @@ function _mosaico_civicrm_alterMailContent(&$content) {
 }
 
 /**
+ * We need to filter URLs from Mosaico when displaying a mailing in the browser (it's already done for sent mails)
+ * The default template social icons get saved as templates/versafix-1/img/facebook.png etc.
+ * We need absolute URLs for display.
+ *
+ * @param $params
+ * @param $context
+ */
+function mosaico_civicrm_alterMailParams(&$params, $context) {
+  // When sending mail, context = flexmailer
+  if ($context == 'civimail') {
+    $filter = new CRM_Mosaico_UrlFilter();
+    $html = $filter->filterHtml(array($params['html']));
+    $params['html'] = reset($html);
+  }
+}
+
+
+/**
  * Implements hook_civicrm_mailingTemplateTypes().
  *
  * @throws \CRM_Core_Exception
--- a/mosaico.php
+++ b/mosaico.php
@@ -361,9 +361,11 @@ function mosaico_civicrm_alterMailParams(&$params, $context) {
     // Convert mosaico tokens to civicrm
     _mosaico_civicrm_mosaicoToCiviTokens($params['html']);
     // Replace tokens with real values
-    $html = _mosaico_civicrm_replaceTokens($params['html'], TRUE);
-    if ($html) {
-      $params['html'] = $html;
+    if (isset($params['html'])) {
+      $html = _mosaico_civicrm_replaceTokens($params['html'], TRUE);
+      if ($html) {
+        $params['html'] = $html;
+      }
     }
 
     // Now filter URLs to make sure they are all valid
--- a/mosaico.php
+++ b/mosaico.php
@@ -269,6 +269,54 @@ function _mosaico_civicrm_check_dirs(&$messages) {
 }
 
 /**
+ * Mosaico templates have a few of their own tokens which are named differently from
+ * CiviMail tokens. By treating these as aliases, we can get more compatibility between
+ * Civi's delivery system and upstream Mosaico templates.
+ *
+ * @param $content
+ */
+function _mosaico_civicrm_mosaicoToCiviTokens(&$content) {
+  $tokenAliases = array(
+    // '[profile_link]' => 'FIXME',
+    '[show_link]' => '{mailing.viewUrl}',
+    '[unsubscribe_link]' => '{action.unsubscribeUrl}',
+  );
+  $content = str_replace(array_keys($tokenAliases), array_values($tokenAliases), $content);
+}
+
+/**
+ * Replace civimail tokens
+ * FIXME: This has already been run once when displaying an email in browser.
+ *   But there were still mosaico tokens present which didn't get parsed.
+ * This is pretty much a direct copy from CRM_Mailing_Page_View::run
+ */
+function _mosaico_civicrm_replaceTokens($content, $html = TRUE) {
+  $mailingID = CRM_Utils_Request::retrieve('id', 'String', CRM_Core_DAO::$_nullObject, TRUE);
+  $contactID = CRM_Core_Session::getLoggedInContactID();
+
+  $mailing = new CRM_Mailing_BAO_Mailing();
+  $mailing->id = $mailingID;
+
+  if (!$mailing) {
+    return FALSE;
+  }
+
+  if (!$mailing->find(TRUE)) {
+    CRM_Utils_System::permissionDenied();
+    return NULL;
+  }
+
+  $domain = CRM_Core_BAO_Domain::getDomain();
+
+  $tokens = CRM_Utils_Token::getTokens($content);
+  $content = CRM_Utils_Token::replaceSubscribeInviteTokens($content);
+  $content = CRM_Utils_Token::replaceDomainTokens($content, $domain, $html, $tokens);
+  $content = CRM_Utils_Token::replaceMailingTokens($content, $mailing, NULL, $tokens);
+
+  return $content;
+}
+
+/**
  * Convert dyanmic-y image URLs to static-y URLs.
  *
  * This is analogous to alterMailContent, but we only apply to Mosaico mailings.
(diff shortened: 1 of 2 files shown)


mattwire:
This doesn't quite work yet for "View in Browser" as the mailing is saved to the DB with API Mailing.create() and the URLs are relative.  I can't find where the call to Mailing.create actually happens - any idea?  It really needs saving to the DB with absolute URLs I think.

mattwire:
Have now pushed changes which allow the "View in Browser" to work properly.  It's a bit of a hack so probably needs some looking at before merging.  It works, and is in use on two of my clients sites.

mosaico.php
```diff
@@ -308,6 +347,36 @@ function($matches){
...
+ */
+function mosaico_civicrm_alterMailParams(&$params, $context) {
+  // When sending mail, context = flexmailer
+  if ($context == 'civimail') {
+    // In this context we are displaying in browser, does it do anything else too?
+    // Convert mosaico tokens to civicrm
+    _mosaico_civicrm_mosaicoToCiviTokens($params['html']);
+    // Replace tokens with real values
+    if (isset($params['html'])) {
+      $html = _mosaico_civicrm_replaceTokens($params['html'], TRUE);
```
totten, line 365: @mattwire You commented on Mattermost about how you weren't "happy with the way some of the hooks in mosaico.php were effectively being double-called". I share the view that it's odd (and possibly problematic).

### Background

The following type of interaction is generally suspicious from a security perspective:

```php
$text = $originalTemplate;
$text = replaceSomeTokens($text);
$text = replaceMoreTokens($text);
echo $text;
```

The problem is that the token substitution from `replaceSomeTokens()` returns text that is re-interpreted by `replaceMoreTokens()`. So the first substitution could output control-codes which cause tricky things in the second one. (This was most dramatically demonstrated circa Civi 2.x... when an end-user could put arbitrary Smarty+PHP code in their `first_name`... and CiviMail executed it. But you might also get tricky situations with `contact.hash` or custom fields.) 

So whenever I see something that looks like iterative substitution, I get squeamish.

`civicrm-core` has (as far as we know) been relatively safe despite breaking this rule -- because it had an order of substitutions which seemed difficult to exploit in a meaningful way. (At least, it did some point in time.) But a major goal in FlexMailer is get away from iterative substitution. Instead, the substitution should be as *atomic* as possible.

### Applied to Mosaico tokens

The `_mosaico_civicrm_mosaicoToCiviTokens()` stuff is safe because the substitutions aren't influenced by user data. `[show_link]` always becomes `{mailing.viewUrl}`.  However, the `mosaicoToCiviTokens` stuff needs to run *before* the main substitution.

My general recollection is the delivery process a bit like this pseudo-code:

```
1: $body_html = '<html>...Hello {contact.display_name}...</html>';
2: CRM_Utils_Hook::alterMailContent(...$body_html...); // Hook for **general content**.
3: foreach ($recipients as $recipient) {
4:  $mailParams = array(...);
5:  $mailParams['body_html'] = evaluateTokens(... $body_html, $recipient ...);
6:  CRM_Utils_Hook::alterMailParams($mailParams); // Hook for **single, concrete message**.
7:  send($mailParams);
8: }
```

The idea is that part of step `#2` is to do `mosaicoToCiviTokens()` (translating the tokens (`[show_link]` => `{mailing.viewUrl}`). Then we can sit back let step `#5` handle the real substitution work.

### Applied to this patch

If I'm reading correctly, this patch taps into step `#6` and does an extra round of `mosaicoToCiviTokens()` and then `evaluateTokens(...)`.

This is problematic if you consider an email sent to the contact `Adam B. {mailing.id}`, wherein:

 * The initial template starts as: `<html>...Hello {contact.display_name}...</html>`
 * Step `#5` expands the token, and the message `<html>...Hello Adam B. {mailing.id}...</html>`
 * Step `#6` fires the hook, and then Mosaico's `hook_civicrm_alterMailParams` evaluates this new token. This produces `<html>...Hello Adam B. 1234...</html>`.

(Is `mailing.id` generally sensitive data? No. Am I confident that the extra substitution is safe in every organizational context / deployment -- and will remain safe with future patches? No. And I don't want to have to think about it!)

But the fact that you had to this extreme feels pretty meaningful -- it suggests that something in the "Preview" operation is not abiding by the expected sequence of operations (from the pseudocode).

mattwire:
@totten That sounds about right.  Note that the first commit (23rd August) I'm completely happy with and think it is correct and sane.  (From memory) that fixes preview and mail-client views.

It's the commits on the 24th which fiddle with the hooks to make "View in Browser" display properly.  I couldn't find a way to get that working without double calling like I've done and I suspect it might need some tweaks in core as well as the extension to let it work properly.  (From memory again..) I *think* the problem comes down to the fact that we are storing mosaico style tokens in the database, which if we weren't the "View in Browser" would work without all the extra hookery.
I didn't want to make any changes to db storage etc. in this patch as that would have created compatibility issues with the "official" civi mosaico release that could be problematic for upgrades later when the "official" civi mosaico extension catches up :-)

dtw:
Half a fix so that the icons at least work in the mailings, even if not in the browser, would be great.

monishdeb:
@mattwire confirmed that your patch fixed the issue, now I can see the social icons rather then text. yayee :D

totten:
@mattwire @monishdeb OK, stepping back a second, I have a more basic issue -- I can't replicate the problem that this PR aims to solve.

For #224, I've gone through the full procedure of [Manual Tests: Images and Links](https://github.com/veda-consulting/uk.co.vedaconsulting.mosaico/blob/2.x/TESTING.md) on my local environment:
 * __Server environment__: PHP v5.6, MySQL v5.7, Apache v2.4, MailCatcher
 * __Codebase__: `dcase` configuration -- Drupal v7.56, CiviCRM `master` (~v4.7.29), Flexmailer v0.2-alpha5, Mosaico v0.15-civicrm-2. (Note: Also enabled CiviEvent so I could use it for an example URL. All extensions are stored under `sites/all/modules/civicrm/ext`.)
 * __Web browser__: Firefox 54
 * __Mail client__: MailCatcher web UI

Everything passes, except for a different issue. (The `{action.forward}` page doesn't work in SC-1 or SC-4. I think the problem is specific to that token/page, and I'd hazard a guess that the problem originated in traditional CiviMail.) 

So I'm wondering ... what's the difference? 
 * Maybe... old version of Flexmailer?
 * Maybe... the mailing HTML includes some relative links?
 * Maybe... the mailing HTML was originally generated by an older version of Mosaico?
 * Maybe... there's a different base-template (with some subtle difference from `versafix-1`)?
 * Maybe... Mosaico produces different output in different browsers (Firefox vs ??)?

totten:
Some more followup on reproducing -- I mentioned this to @bgm, and he said that `civicrm.org` had experienced the same problem. So I did a more real-world test:

 * Deployed the latest Flexmailer, Shoreditch, and Mosaico on `civicrm.org`
 * Created a new mailing (`versafix-1`) with an uploaded image with the social media footer.
 * Checked all scenarios (SC-1,2,3,4). Both uploaded images and social media images rendered correctly in Gmail and Mac Mail.

mattwire:
@totten Just added comments to #160 - still seems to be broken :-(  At least for the built-in html preview - let me know if you are able to reproduce (that part should be fixable by the filterHTML() function changes in this PR).

totten:
@mattwire - Hrm. I'm still struggling to reproduce. FWIW, I've tried more permutations with:
 * CiviCRM `4.7.27` and `4.7.29`
 * Mosaico (upstream) `v0.15-civicrm-2` and `master-civicrm-1-mjwconsulting` (Note: running `npm install` and `cv flush` after a change)
 * Firefox 57, Chrome 62, and Safari 11
 * Drupal 7.56 and WordPress 4.8.1

FWIW, when I use the built-in HTML preview and inspect the IFRAME source code, it shows the correct absolute URL.

screen shot 2017-12-04 at 2 42 31 pm

__Question__: What URL(s) does your system produce for the social icon(s)?

Interestingly, on my system, *none of the URL filtering* in `UrlFilter.php` or `_mosaico_civicrm_alterMailContent()` is necessary for social icons. I can comment out all the server-side image filtering (`preg_replace_callback` lines in `UrlFilter.php` and `_mosaico_civicrm_alterMailContent()` ), and the "Preview as HTML" still produces the proper, absolute URLs for social icons. 

Which makes me wonder -- is the server-side really supposed to be responsible for this? 

If one uses the demo at https://mosaico.io/ and hits "Download", it spits out HTML with correct, absolute URLs. (That's primarily a client-side app with client-side storage...) In reading the source-code for [other example backends](https://github.com/voidlabs/mosaico/wiki/Serving-Mosaico), I can see that they filter *uploaded images* (`/img?src=`) -- but I can't see the server-side doing anything to absolute-ify the social icons.

If converting social icons to absolute URLs is actually a client-side function, then maybe it keys off something that's incorrect/buggy? For example, maybe the `<BASE HREF="...">` is wrong or missing?

__Example__: On my local Drupal site, the editor IFRAME points to `http://dcase.l/civicrm/mosaico/iframe?snippet=1`. The full source is https://gist.github.com/totten/bbdfa360e79880bb1739e62326b6c282 

Perhaps you could post a similar dump?

totten:
The comments on #160 (esp starting around [here](https://github.com/veda-consulting/uk.co.vedaconsulting.mosaico/issues/160#issuecomment-358537768)) indicate that the problem is in the data -- i.e. that templates/designs predating beta3 have problems (and carry-forward when re-used), but that templates/designs created after beta3 work OK without this patch.

* _If it's true_ that the problem is hold-over data from pre-beta3:
    * It would be useful to provide admins with some kind of report/detection. (E.g. Subscribe to `hook_civicrm_check()` and display a warning if the DB has any problematic templates). However, arguably it's not release-critical (since it affects beta-period data).
* _If it's not true_ that the problem is hold-over data from pre-beta3:
    * Please feel free to reopen and share more details.

JoeMurray:
FWIW, we found that the change on L294 https://github.com/veda-consulting/uk.co.vedaconsulting.mosaico/pull/182/files#diff-d8205fe762b98a885815f3d4bb175886R294 fixed a mailing issue breakage https://civicrm.stackexchange.com/questions/23047/send-schedule-mailings-job-succeeds-but-scheduled-mailing-not-completely-sent/23138#23138

PR #190, opened 2017-09-25
Fix Mosaico menu covered by Drupal toolbar

eruraindil:
Problem: in our Drupal site we use the User Shortcuts which stick down below the CiviCRM menu, but this z-fights with the Mosaico menu.

image

This PR fixes the issue by checking for the existence of the Drupal toolbar and offsetting the Mosaico iframe by whichever is larger.

image


brucew2013:
Sorry I misunderstood. Got it now. Thanks.

PR #199, opened 2017-10-17
Fix for menu covered by Joomla menu

mattwire:
Based on PR #190 but for Joomla!

### DIFF
--- a/ang/crmMosaico.ang.php
+++ b/ang/crmMosaico.ang.php
@@ -25,6 +25,7 @@ $result = array (
     // in these jQuery selectors.
     'topNav' => '#civicrm-menu',
     'drupalNav' => '#toolbar',
+    'joomlaNav' => '.com_civicrm > .navbar',
     'leftNav' => '.wp-admin #adminmenu',
   ),
 );
--- a/ang/crmMosaico/Iframe.js
+++ b/ang/crmMosaico/Iframe.js
@@ -28,6 +28,9 @@
           if (c.topNav && $(c.topNav).length > 0) {
             if (c.drupalNav && $(c.drupalNav).length > 0 && $(c.drupalNav).height() > $(c.topNav).height()) {
               top = $(c.drupalNav).height();
+            }
+            else if (c.joomlaNav && $(c.joomlaNav).length > 0 && $(c.joomlaNav).height() > $(c.topNav).height()) {
+              top = $(c.joomlaNav).height();
             } else {
               top = $(c.topNav).height();
             }


totten:
@mattwire I've tested/reviewed. Mostly 👍  with one suggested change.

* ([`r-code`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-code)) __Pass__: The code is a clear extension of similar code for Drupal/WP.
* ([`r-doc`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-doc)) __Pass__
* ([`r-maint`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-maint)) __Pass__
* ([`r-run`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-run)) __Pass__:
    * I tried this patch in both a Joomla build (`civihydra create https://download.civicrm.org/latest/civicrm-RC-joomla.zip`) and Drupal build (`dcase`) with very recent code.
    * For Joomla, the patch fixes misalignment of the navbar. 👍 
    * For Drupal, it worked the same before. 👍 
* ([`r-user`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-user)) __Pass__: Given that Mosaico is trying to avoid collision with the main menus in WP+Drupal... it makes sense to do the same on Joomla.
* ([`r-tech`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-tech)) __Issue__:
    * The WP selector (`.wp-admin #adminmenu`) is a better example to emulate than the Drupal selector (`#toolbar`). 
    * Why? The selector is *the test* used by Mosaico to figure out which CMS is active. The JS logic walks through each of these selectors and picks the first one that has a match. But if you consider a range of current/future environments, merely having an element that matches `.navbar` or `#toolbar` doesn't seem like a very strong signal about which CMS is active. Including something distinctive to the CMS should avoid confusion.

For Joomla, I think "`.com_civicrm .navbar`" or "`.com_civicrm > .navbar`" would be better.

```diff
diff --git a/ang/crmMosaico.ang.php b/ang/crmMosaico.ang.php
index fc7d81b..2a4e657 100644
--- a/ang/crmMosaico.ang.php
+++ b/ang/crmMosaico.ang.php
@@ -25,7 +25,7 @@ $result = array (
     // in these jQuery selectors.
     'topNav' => '#civicrm-menu',
     'drupalNav' => '#toolbar',
-    'joomlaNav' => '.navbar',
+    'joomlaNav' => '.com_civicrm > .navbar',
     'leftNav' => '.wp-admin #adminmenu',
   ),
 );
```

mattwire:
@totten Agreed!  I've updated the PR with your suggested change to the selector.

totten:
Thank you, @mattwire!

Issue #202, opened 2017-11-03
Mosaico only shows 25 templates


pbatroff:
In the template overview (civicrm/a/#/mosaico-template) a maximum of 25 templates are shown. If more templates are created, they temporarily show up (and correctly), but once the browser cache is cleared they just disappear. There are no error logs or interaction in the UI.

This happens when the templates are pulled from the Database, the default limit of elements is 25. Other templates are not shown.

A good long term solution would be to implement a a pager for a bigger amount of templates, but for now it would be best to just pull all available templates. The limit (correctly) needs to be set to zero.

pbatroff:
I added a PR with a proposed solution with option limit set to 0 (PR #203)

totten:
Resolved by #203.

PR #203, opened 2017-11-03
Correct syntax for option limit when getting templates from db (#202)

pbatroff:
This fixes issue #202. The option limit is now set correctly

### DIFF
--- a/ang/crmMosaico/Templates.js
+++ b/ang/crmMosaico/Templates.js
@@ -29,8 +29,8 @@
     }
 
     crmApi({
-      bases: ['MosaicoBaseTemplate', 'get', {sequential: 1, limit: 0}],
-      templates: ['MosaicoTemplate', 'get', {sequential: 1, limit: 0, return:['title', 'base']}]
+      bases: ['MosaicoBaseTemplate', 'get', {sequential: 1, 'option.limit': 0}],
+      templates: ['MosaicoTemplate', 'get', {sequential: 1, 'option.limit': 0, return:['title', 'base']}]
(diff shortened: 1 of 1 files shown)


totten:
@deepak-srivastava I think this is good to merge. Here's more a detailed review:

* ([`r-code`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-code)) __Pass__
* ([`r-doc`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-doc)) __Pass__
* ([`r-maint`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-maint)) __Pass__
* ([`r-run`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-run)) __Pass__:
    * First, I created a bunch of `MosaicoTemplate`s using a shell command like:
      * `for n in $(seq 1 28) ; do cv api MosaicoTemplate.clone id=4 title="Template #${n}" ; done`
    * Navigated to "Mailings => Mosaico Templates" and confirmed that it was missing several templates (because of the cutoff)
    * Applied the patch. Cleared cache.
    * Relaoded "Mailings => Mosaico Templates" and confirmed that all templates were now present.
* ([`r-user`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-user)) __Pass__: The PR's new behavior is how I'd expect the screen to work.
* ([`r-tech`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-tech)) __Pass__: Only affects GUI users -- shouldn't affect API or data model.

PR #204, opened 2017-11-06
README.md - Changes to text

AnneDru:
Hi, I'm Anne. During the CiviSprint I was tasked with looking at the Extension Readme's from a 'non-technical user' point of view, so that the Readme would contain all of the relevant information and would use language that a non-tech user would understand. In other words, I would be more inclined to install an extension if I understood what it was about and had enough info to be able to install it.  Clearly some extensions would require more technical support, but the Readme would indicate that.

Also, the various extensions would follow a similar format. To this end a boilerplate is being set-up for extensions developed in the future.

I hope this makes sense.  So, this is my first go at looking at the top 20 extensions and making suggestions to start this project.  So you are my 'trial' and I hope you don't mind. I welcome any feedback on a) my technique and b) the actual suggested changes to your Readme.  Which, by the way, I think is really excellent well presented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,9 @@
 # uk.co.vedaconsulting.mosaico
+
+##Extension title 
+Integration of Mosaico with CiviCRM
+
+##Description
 This extension integrates Mosaico a responsive email template editor, with CiviCRM.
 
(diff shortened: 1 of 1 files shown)


totten:
@deepak-srivastava This seems pretty safe.

PR #206, opened 2017-11-14
mosaico.civix.php - Update template

totten:
This just updates the boilerplate codd in `mosaico.civix.php` to match the
latest civix template.  The main change is that it adds the big `E` helper
class, which provides a prettier way to lookup the name/path/url of the
extension.


totten:
I'm working on some other PR's which require generating more code -- and they rely on having the current template. Merging.

PR #210, opened 2017-11-22
Mosaico.setting.php - Switch default layout back to single-page

totten:
## Background

Early mockups for Mosaico 2.x proscribed a BootstrapCSS-based wizard layout. However, during implementation, we had to address these bits incrementally -- e.g. first, incorporating BootstrapCSS; then developing a Bootstrap-based wizard component; etc. This led to an accidental discovery: an intermediate revision had a single-page layout, and it actually felt pretty easy to use. To see if this was just a personal fluke, I showed both layouts to a couple outsiders (non-Civi folks) who had the same reaction.

Unfortunately, that's anecdotal. There's still a strong element of subjectivity (personal taste, organizational context), and it would be more informative to check with a range of users. After the initial discovery, Jamie agreed to include the layout as a variable in some focus-group user testing, but they ultimately abandoned this. (Guess: the timing didn't work out.) Thus, the debate between "single page layout" and "wizard layout"  essentially remained a matter of irresolvable subjectivity.

## Demo Day: About

I now have more meaningful data with which to address the question of default layout.

Each year, there's a conference in Oakland for non-profit activists/technologists/developers. It attracts people from a number of small and mid-size NGOs (mostly in US; some Latin America), and it includes people with a range of experience in CiviCRM, Salesforce, CiviMail, MailChimp, Constant Contact, etal. It's fairly diverse -- some people have deep experience in several tools; some are just evaluators; many are users; others are consultants or developers. All are in the target market for CiviCRM/CiviMail.

As part of the conference, we had a "demo day" exercise. The format was basically this: you get a group of 5-8 people huddled around a computer, and you have 8 minutes to show off a demo.  Then the audience rotates, and a new group of 5-8 people comes in.  There are 5 iterations of this, which means you get 25-40 people in total.

I used the "demo day" to show off a bunch of the Civi improvements -- e.g. Shoreditch's "View Case", e.g. Mosaico's "New Mailing", e.g. the new CiviCase. *All reactions to the new revisions were quite positive, especially from people who had previously used the old screens.* I think we had ~35 people.

Of course, we've also had this open question about layout... and there's no empirical data... and I've been genuinely curious to understand why reactions differ so strongly. Demo Day seemed like a good opportunity to get real feedback from a large group of users. So, after showing the goodies, I used the last 2 minutes to say: "the development team had some debate about one aspect of Mosaico, and we'd like to get your feedback..." Then I'd demo both layouts and ask for reactions/preferences.

This process didn't have the rigor of a real focus group -- eg I didn't have time to take detailed, contemporaneous notes. The questions weren't planned out in advance, and there wasn't time for much follow-up. But I did capture notes within a day, and I tried to reduce biases in the presentation -- changing the order of presentation, avoiding loaded statements/questions, etc.

IMHO, despite these limits, it's the best data we have so far.

## Demo Day: Results

* Generally:
   * I didn't have time to record exact numbers.
   * Users seemed to prefer the single page layout while consultants seemed to prefer the wizard. 
   * The overall tilt was unmistakable. The ratio of single-vs-wizard preferences was probably 2:1 or 3:1.
* Here a few comments which seemed insightful:
   * __Comment from a consultant, paraphrased__: "When supporting a client, I'd want them using the wizard so that they could tell me exactly which step they were on."
  * __Comment from the same consultant, paraphrased__: "But I wouldn't want to use the wizard myself."
  * __Comment from a user, paraphrased__: "I fiddle a lot with the subject and recipients while working on the mailing. The wizard would make that hard."
  * __Comment from a user, paraphrased__: "The single page is different from other email software (MailChimp, Constant Contact), but it looks easier."
* With one group, I did a follow-up question. The original question was something like, "Which layout seems better?"  The group was originally mixed (~2 for wizard and ~4 for single-page), but the follow-up question shifted the emphasis: "What would you prefer to *use yourself*?" And the result became unanimous: they all preferred to *use* the single-page.

## Comments

As a developer or designer sitting at a computer and working with Github+code, the difference between layouts seems a bit arbitrary and ambiguous. However, *the experience of actually showing each screen to a large pool of users* makes it more visceral. The available information indicates that more users will feel satisfied with usability if the "New Mailing" screen defaults to a single-page layout.

This PR changes the default accordingly.

### DIFF
--- a/settings/Mosaico.setting.php
+++ b/settings/Mosaico.setting.php
@@ -13,7 +13,7 @@ return array(
     'pseudoconstant' => array(
       'callback' => 'CRM_Mosaico_Utils::getLayoutOptions',
     ),
-    'default' => 'bootstrap-wizard',
+    'default' => 'auto',
     'add' => '4.7',
     'title' => 'Mosaico editor layout',
     'is_domain' => 1,


totten:
@mattwire  @deepak-srivastava -  Matt started drafting some release-notes, so I want to point out that the titles/descriptions on #210 and #236 don't quite paint an accurate picture of what happens now.

The actual behavior is this:

 * The default value of `mosaico_layout` is `auto`.
 * On traditional sites, `auto` means "single-page layout with `crm-*` CSS."
 * On Shoreditch sites, `auto` means "wizard layout with Bootstrap CSS."
 * See also: https://github.com/veda-consulting/uk.co.vedaconsulting.mosaico/blob/8eb6f175015972c1335544e655a9ae77044aacb7/CRM/Mosaico/Utils.php#L49-L56

The PR history is a bit misleading because there was slow-motion, under-the-hood back-and-forth around the definition/usage of `mosaico_layout=auto`.

If we want to thoroughly make the policy change described in #210, then we'd need a small patch to the `auto` logic. OTOH, I'm not really pushing for that -- I know Jamie's a big fan of wizard layout, and maybe the status quo is an easy compromise (wherein Shoreditch and `crm-*` have different defaults).

jamienovick:
@totten a change of this magnitude should have the appropriate review. I'm also concerned that were reverting weeks of work that has been paid for by a client to design, style and test a solution that is being used in the wild by several thousand sites. Also note that this will break Shoreditch tests so has knock on effects. I'd add that the designs were also subjectively tested with the opposite results with the paying clients - people who are actually using this solution.

Reading the above it's clear that the results are subjective and possibly biased. A few points:

1. Putting a screen in front of people and asking then to chose is not a great way of gathering meaningful feedback. We should be asking users to perform a task and seeing how quickly they can achieve it and/or where they get stuck. This is an objective measurement. People are very bad at knowing what is 'best'.

2. I'd question whether the people in the room were the right people to ask. The people in the room at a CiviCRM related event are largely engaged or existing users and/or consultants and implementors. We should be testing this with an audience of new users of the CRM, or those who have never used it before and specifically those who are coming from other mailing applications such as MailChimp.

Our feedback from training sessions with users is that they want wizards/simplified screens with clear logical steps that make them confident that they are doing things in the correct order. Too many options on a screen makes people lose confidence. We haven't had any negative feedback on the wizard layout at all... ever!

Users coming from MailChimp and other applications are used to the wizard approach.

If we want to do some proper testing of this in order to validate whether a change is required can I suggest a different approach?

We should make it easy in the CiviMail component settings to select between the screens. We should then gather pingback statistics on which people do pick and base our decision on a larger sample size of actual users. In an ideal world we would probably have a layout selected at random on install and see what the amount that 'switch' are. After a period (say 6 months) we evaluate the data and pick a final solution.

Short of that I would suggest that we stick with what we have and notify users that they can select an option when they first instal letting them make an informed choice.

I have serious concerns about unilaterally changing the UX of CiviCRM for such a major component at this stage and as such I would argue that we need to get appropriate data to support such a change.

totten:
The titles/descriptions on #210 and #236 don't paint an accurate picture of what's happened.

The actual behavior is this:

* The default value of `mosaico_layout` is `auto`.
* On traditional (non-Shoreditch) sites, `auto` means "single-page layout with `crm-*` CSS."
* On Shoreditch sites, `auto` means "wizard layout with Bootstrap CSS."

*In other words*, the feature you paid for is still there by default for anyone on Shoreditch... which would be basically everyone currently using 2.0 and everyone you care about. Nothing has been ripped out. And Mosaico still promotes Shoreditch.

Aside from confusing Github data, I don't see what the problem is.

> A sample size of 6 is definitely not large enough to make this change... You are also changing the UX for a large number of installs without any due reason.

Huh? You make it sound like #210 was just asking 6 people. It was **~5 rotations of 5-8 people each**, putting it in the ballpark of ~35 people. That's like running 5 informal focus-groups. Granted, the methodology is not rigorous by professional marketing standards, but it's a lot better than par around here, so it would be a pretty good reason to make a change. I'd love it if we could administer these kinds of things more often for more questions.

> Also your proposal does t solve the UX issue of having large numbers of templates.

The real solution to large numbers of templates is to update the template-gallery with its own organizational tools (filtering, sorting, grouping). That's orthogonal to this discussion -- an overwhelming template-gallery is still going to be an overwhelming template-gallery no matter what container you put it in.

caracolmzd:
Guys. Add an "advanced" link in the wizard to disable it and access the single-page layout but keep the wizard as the default.

jamienovick:
@ginkgomzd I think it would be an advanced setting on the civimail component settings... but I think Tim is suggesting that single page would be the default if Shoreditch is not installed (although I'm not clear on the mechanics of this quite yet).

@totten Does the order that you install Shoreditch/Enable as to Mosaico make any difference or is the layout selected on the fly?

totten:
We currently have a screen under "Administer => CiviMail => Mosaico Settings":

screen shot 2018-06-13 at 12 09 44 pm

The dropdown corresponds to the setting `mosaico_layout`, and its default value is `auto`. With `auto`, the layout is selected on the fly based on current configuration. It doesn't matter what order the modules are installed -- it only matters what's active/available when you open the editor. So:

* If you open the editor while Shoreditch is enabled, `auto` means "Wizard (Bootstrap CSS)" (aka `bootstrap-wizard.html`).
* If you open the editor while Shoreditch is not enabled, `auto` means "Single Page (crm-*)" (aka `crmstar-single.html`). 
* There is no condition in which `auto` means "Single Page (Bootstrap CSS)" (aka `bootstrap-single.html`).

@ginkgomzd Yeah, I can see how it might be better (or worse) to have the setting inside an "Advanced" part of the mailer UI (rather than a backend/site-wide "Administer" screen). In this case, I think laziness/inertia will win out.

PR #212, opened 2017-11-22
Use .outerHeight() rather than .height() on topNav to include padding

dsnopek:
In order to get the iframe position right the way we're dealing with the 'topNav' on Drupal 8, we need to use `.outerHeight()` rather than `.height()` because we need to count the padding too.

### DIFF
--- a/ang/crmMosaico/Iframe.js
+++ b/ang/crmMosaico/Iframe.js
@@ -26,10 +26,10 @@
           var c = CRM.crmMosaico || {};
           var top = 0, left = 0, width = $(window).width(), height = $(window).height();
           if (c.topNav && $(c.topNav).length > 0) {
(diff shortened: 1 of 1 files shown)


totten:
Thanks, @dsnopek!

I checked out locally, resolved the conflicts, and re-tested in the following environments:
 * `dcase` (Drupal 7 with the `seven` theme; plus Shoreditch provides the main `civicrm.css`)
 * `dmaster` (Drupal 7 with the `garland` theme; Shoreditch enabled, but it's not providing the main `civicrm.css`)
 * `wpmaster` (WordPress 4.8; Shoreditch enabled, but it's not providing the main `civicrm.css`)
 * `hydra-joomla` (Joomla 3.6; Shoreditch enabled, but it's not providing the main `civicrm.css`)

In each cases, the sizing of the Mosaico dialog seemed about as good before. 👍  And I'll take your word that it's better on D8. :)

dsnopek:
Awesome, thanks!

Issue #143, opened 2017-04-26
Requires unsubscribe/opt-out token to send email even though unsubscribe link is already in header


madurak:
Unsubscribe link is already in the header (of the email in template) but error message says it still requires unsubscribe/opt-out token to send the mail.
<img width="1391" alt="screen shot 2017-04-26 at 17 44 56" src="https://cloud.githubusercontent.com/assets/10810117/25446045/1b20040a-2aa8-11e7-9416-92b7b3bfe5f0.png">

petednz:
I think the OP problem is that they haven't inserted {domain.address} token which is also required.

petednz:
My observation is that trying to add eg OptOut via webpage token in to the footer block results in this <domain>/civicrm/mailing/optout?reset=1&jid=1763&qid=1763&h=fakehash showing in both the Preview and the Test.

petednz:
I think the OP is simply failing because they haven't inserted {domain.address} token which is also required.

totten:
I'm able to reproduce this problem.

This is sort of a Mosaico issue, but IMHO it really goes to the concept behind FlexMailer -- *extensions should be able to define new template syntaxes* . Mosaico is a basic example of this (because it adds  different syntax for the unsub tokens), but other use-cases (like Pug, Markdown, Mustache, or Smarty) could also have their own syntaxes for the unsub tokens. I'm going to work a bit toward making the CAN-SPAM validation tweakable through FlexMailer.

totten:
Related: https://issues.civicrm.org/jira/browse/CRM-21472

totten:
Depends: https://github.com/civicrm/org.civicrm.flexmailer/pull/9 and https://github.com/civicrm/civicrm-core/pull/11316

totten:
For posterity: this should be fixed with the combination of  CiviCRM 4.7.29+, Mosaico 2.0-beta3+, and FlexMailer 0.2-alpha5+. If you run on an older version of CiviCRM 4.7.x:

 * The new code should degrade gracefully (e.g. no hard crashes, but same buggy behavior).
 * You can easily work-around by editing your content: simply change `[unsubscribe_link]` to `{action.unsubscribeUrl}`.

PR #215, opened 2017-11-25
Accept Mosaico-style unsubscribe URLs during validation (#143)

totten:
Before
------
As observed in #143, the default `versafix-1` templates use tokens like
`[unsubscribe_link]` which do not match `{action.unsubscribeUrl}`. Thus,
when you try to submit the mailing, it fails because the token is missing.

After
-----
The `[unsubscribe_link]` is treated as an alias for
`{action.unsubscribeUrl}`. Thus, when you try to submit the mailing,
it accepts that token.

Comment
-------
There's another required token, `{domain.address}`. For the moment,
the patch waives that requirement because Mosaico UX handles that
differently. See comments inlined.

This patch depends on a recent update to Flexmailer: https://github.com/civicrm/org.civicrm.flexmailer/pull/9 . That patch is currently merged.

To fully work, this patch also depends on an open core PR, https://github.com/civicrm/civicrm-core/pull/11316 . However, it is safe to use this patch without 11316 -- the new code is simply inert.

### DIFF
--- a/CRM/Mosaico/MosaicoRequiredTokens.php
+++ b/CRM/Mosaico/MosaicoRequiredTokens.php
@@ -20,8 +20,22 @@ class CRM_Mosaico_MosaicoRequiredTokens extends RequiredTokens {
   }
 
   public function getRequiredTokens() {
-    return Civi::service('civi_flexmailer_required_tokens')
+    $requiredTokens = Civi::service('civi_flexmailer_required_tokens')
       ->getRequiredTokens();
+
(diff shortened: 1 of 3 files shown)


seamuslee001:
@totten i would note that the core PR and flexmailer PRs are now both merged, should this get merged then?

totten:
Yeah, I held off a little on merging -- if you deployed the first couple commits on an outdated version of Flexmailer, then it'd be cause a hard crash, so I wanted to mitigate that. It should now degrade more gracefully.

Issue #153, opened 2017-05-13
Hangs after selecting template in Step 2


brucew2013:
OS: CentOS 7.3 x86-64 cPanel/WHM virtuozzo
Webserver: Litespeed 5.1.15
PHP: 7.1 
CMS: Joomla 3.7.0
CRM: CiviCRM 4.7.19

After needing to install Shoreditch and Flexmailer plus ImageMagick and FileInfo PHP modules it began to work.

In step 2, selecting the default template hangs and disables all (button) links in CiviCRM and Mosaico. (See attached; notice upper right corner "loading") and throwing JS error:
> Uncaught ReferenceError: $ is not defined at development.alcoholjustice.org/:42

Opening that up highlights line 42. Is this correct?:
```
41 <script type="text/javascript">
=>42    $(function() { 
43      if (!Mosaico.isCompatible()) {
44        alert('Update your browser!');
45        return;
46      }
```

<img width="733" alt="screen shot 2017-05-13 at 4 14 28 pm" src="https://cloud.githubusercontent.com/assets/5554148/26029301/6badcb6c-37f7-11e7-8172-38713cb7bc0a.png">

brucew2013:
Additionally, in v.2, it looks like the path is incorrect. 
What browser console shows (see below) but doesn't exist in the Git ZIP package:
**JROOT/media/civicrm/ext/uk.co.vedaconsulting.mosaico/packages/mosaico/templates/versafix-1/edres/_full.png**

What exists is:
**JROOT/media/civicrm/ext/uk.co.vedaconsulting.mosaico/templates/CRM**

<img width="1289" alt="screen shot 2017-05-13 at 6 30 15 pm" src="https://cloud.githubusercontent.com/assets/5554148/26030019/8e66d906-380a-11e7-8f6a-f479ed7baef9.png">

stevekessler:
This is also happening during step 1 if you are not using the Shoreditch.
<img width="1792" alt="screen shot 2017-05-14 at 10 17 25 pm" src="https://cloud.githubusercontent.com/assets/867985/26042534/816e0e1a-38f3-11e7-8210-e094d322ed37.png">

rekaeps:
I'm having the same issue in Drupal 7.54 and CiviCRM 4.7.19. Running PHP 7.

smch:
Same problem, Civi 4.7.20

totten:
Background: 
[...]

PR #217, opened 2017-11-25
mosaico_civicrm_check - Warn if dependencies are missing (#153)

totten:
Sometimes people skip the install instructions. Display a warning if they're missing key dependencies.

See also: #153.


seamuslee001:
This looks good and sensible to me i haven't tested yet but the idea makes sense

totten:
Thank you, @seamuslee001. Merging.

PR #219, opened 2017-11-27
Fix display of group listings on newer builds of CiviCRM

totten:
Upstream CiviMail made some changes to how the list of groups is managed, which breaks Mosaico on newer builds.

Note the use of `CRM.crmMailing.testGroupNames || CRM.crmMailing.groupNames` is intended to provide compatibility with older and newer protocols.

Problem reported by @eileenmcnaughton on Mattermost (`dev-mosaico`). @seamuslee001 noted similar bugs+patches in traditional CiviMail.


seamuslee001:
Changes look right to me

PR #225, opened 2017-11-30
Migrate templates from 1.x to 2.x

totten:
Overview
---------
This PR adds a migration assistant to help folks move templates from v1.x to v2.x

It also updates the schema so that we can preserve the data about msg_tpl_id mappings. The data isn't used within this PR -- but it's likely to be used by an experimental extension.

The "migration assistant" approach is a little less automated than using `updated_NNNN()`, but it allows us to cope more transparently with the oddball scenarios that evaluators are likely to have created when juggling v1.x and v2.x.

Before
-------
Templates from v1.x and v2.x live in separate tables. 

When you upgrade to 2.x, your old templates are no longer available for editing via Mosaico.

After
-----
If you have templates from v1.x, a status alert is displayed:

screen shot 2017-11-29 at 11 49 53 pm

Which directs you to a migration assistant:

screen shot 2017-11-29 at 11 49 36 pm

The migration tasks are also available as APIs (`Job.mosaico_migrate` and `Job.mosaico_purge`).

### DIFF
--- a/CRM/Mosaico/DAO/MosaicoTemplate.php
+++ b/CRM/Mosaico/DAO/MosaicoTemplate.php
@@ -116,6 +116,12 @@ class CRM_Mosaico_DAO_MosaicoTemplate extends CRM_Core_DAO {
    */
   public $content;
   /**
+   * FK to civicrm_msg_template.
+   *
+   * @var int unsigned
+   */
+  public $msg_tpl_id;
+  /**
    * class constructor
    *
    * @return civicrm_mosaico_template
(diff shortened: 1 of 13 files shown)


totten:
To test the upgrade/migration path, I went through this procedure:


* Use v1
    * Install `mosaico` extension v1.x
    * Create two templates
    * Observe: Two message templates exist
    * Send an individual message with the template
* Use v2
    * Install `mosaico` extension v2.x
    * Create a third template
    * Run the Migration Assistant's "Copy".
    * Observe: The original two templates are imported. Now we have a total of three templates.
    * Create a mailing. For each of the three templates:
        * Reset the mailing's design. Pick template `$X`. Preview in browser. Send test email.
* Use message-template sync
    * Install `mosaicomsgtpl` extension
    * Observe: Two pre-existing message templates. One new.
    * Edit a mosaico template. Observe: message template updates.
    * Send an individual template with one of them.

PR #226, opened 2017-12-01
Remove unused `civicrm/mosaico/dl`

totten:
In v1.x, this end-point was used for saving templates and sending test
messages.  In v2.x, it's been replaced by APIv3 calls.


totten:
To ensure that v2.x really wasn't calling this interface, I've:

 * Grepped the codebase for `processDl` and `civicrm/mosaico/dl`
 * Re-tested the "Mosaico Templates" admin screen, and attached images in various ways.
 * Re-tested the "New Mailing" screen, and attached images in various ways.

PR #227, opened 2017-12-01
Add custom templates directory and url settings

francescbassas:


CRM/Mosaico/BAO/MosaicoTemplate.php
```diff
@@ -38,20 +38,33 @@ public static function findBaseTemplates() {
 
       $templatesUrl = CRM_Mosaico_Utils::getTemplatesUrl('absolute');
 
+      $templatesLocation[] = array('dir' => $templatesDir, 'url' => $templatesUrl);
+
+      $customTemplatesDir = CRM_Core_BAO_Setting::getItem('Mosaico Preferences', 'mosaico_custom_templates_dir');
+      $customTemplatesUrl = CRM_Core_BAO_Setting::getItem('Mosaico Preferences', 'mosaico_custom_templates_url');
+      if (!is_null($customTemplatesDir) && !is_null($customTemplatesUrl)) {
+          if (is_dir($customTemplatesDir)) {
+            $templatesLocation[] = array('dir' => $customTemplatesDir, 'url' => $customTemplatesUrl);
```
totten: If you have a site that goes through migrations (e.g. dev=>prod; prod=>dev; prod to a different domain), then storing an absolute path or absolute URL makes the migration more difficult.

We've started storing paths which are relative to some base, such as `[civicrm.root]/foo` or `[civicrm.files]/bar` or `[cms.root]/foobar`. The `[...]` are evaluated at runtime. This improves portability, and it makes it easier to set default values. (A default like `/var/www/files/mosaico_tpl` is not portable, `[civicrm.files]/mosaico_tpl` would be portable.)

The smallest change would be to pass the settings through `Civi::paths()`, e.g.

```php
$templatesLocation[] = array(
  'dir' => \Civi::paths()->getPath($customTemplatesDir), 
  'url' => \Civi::paths()->getUrl($customTemplatesUrl),
,); 
```

An absolute path or URL will get passed-through, and relative path or URL will have the appropriate substitution.

mattwire:
This is a logical extension of the work I did to support dynamic templates and has been requested by a number of people.  It simplifies upgrades for those using custom templates and removes the requirement for a writeable extensions directory.
@totten If you are happy with this, I'd say merge.  Is there anywhere we can document this?

totten:
Agree it's a good idea.

A few updates:
 * While testing, I found that the calls to `Civi::paths()` needed to be a bit earlier (before the `is_dir()` test). So I pushed that up (43514e69436bb422d8ec146dc277bf942c552fad). 
 * Added a hook interface -- so you have more options for how to package the template (e.g. as Civi extension). (bc46ffc83fc3e7f62746e1506346270765fa731b)
 * For docs, it'd be really great to reorganize the Markdown files and setup a nicer "book" with mkdocs. For the moment, I just setup a file `API.md` with some notes about these changes as well as some other APIs. (8e274f994302d158537811b899f9c9f428f3e8a0)

PR #232, opened 2018-01-05
CRM-19252 Update Mosaico Base upgrader so Extension versions are inde…

seamuslee001:
…pendant

ping @totten @deepak-srivastava 

This ensures that mosaico upgrader sets the schema version in civicrm_extension not civicrm_setting so its domain independent see CRM-19252. I note that mosiaco.php is already calling the post install hook


mattwire:
@totten This looks pretty safe / standard as it's just updating an auto-generated file

totten:
Thanks, @mattwire @seamuslee001. Agree it looks pretty standard, and it's probably better to get this in before 2.0-final.

PR #249, opened 2018-04-27
CRM-21100 Alter loading of unsubscribe groups and preview groups to be dynamic …

seamuslee001:
…to reduce DOM Load and add performance improvements

This aims to remove some custom handling for groups and to replicate changes made here https://github.com/civicrm/civicrm-core/pull/10898/files that have been accepted into core. 

ping @totten

### DIFF
--- a/ang/crmMosaico.bootstrap/BlockPreview.html
+++ b/ang/crmMosaico.bootstrap/BlockPreview.html
@@ -32,15 +32,11 @@ Vars: mailing:obj, testContact:obj, testGroup:obj, crmMailing:FormController
 
 <div class="form-group">
   <label for="preview_test_group">{{ts('Send test email to group:')}} <a crm-ui-help="hs({id: 'test', title: ts('Test Email')})"></a> </label>
-  <select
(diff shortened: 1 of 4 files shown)


deepak-srivastava:
@seamuslee001 seems like one file is in conflict -  ang/crmMosaico/BlockPreview.html 

@totten any comments?

seamuslee001:
@deepak-srivastava have fixed the conflict now i believe

mattwire:
@seamuslee001 Quick question.. does this add a minimum core version dependency?

seamuslee001:
@mattwire i don't believe so but its just about making sure `crm-entityref` is there which i think is certainly there in the 4.7 series but i basically worked off code. This code should be making it less dependent on core by using more API than preloaded objects @colemanw

colemanw:
This looks right to me :)
@mattwire `entityRef` is also in 4.6 so this shouldn't change the version requirements.

mattwire:
@totten @deepak-srivastava So I think this one can be merged...

PR #250, opened 2018-05-02
TinyMCE editor plugin settings

sunilpawar:
Allow admin to add/remove tinymce editor plugin.

domain-name/civicrm/admin/mosaico?reset=1
mosoico_admin


mattwire:
@sunilpawar can we have some context help on that page because I've no idea what would happen if I changed those new settings

sunilpawar:
@mattwire we have description, can i add more details in description itself or add help icon in-front of field , that will open help text in top right corner ?

mattwire:
@sunilpawar I was thinking that a "help icon" in front of field would be good, because I think you can do many things by changing these new fields, including breaking mosaico completely... Maybe also a "reset to defaults" would be helpful?

totten:
On a technical level, this code looks like a faithful implementation of its design. It's nice to read.

On a design level, I'm still digesting a bit.

* In the FOSS space, clients/consumers can bring strong opinions about things like toolbars/plugins. Exposing this for customization is nice as a way to resolve/avoid conflicts over something that we really don't care to argue about.
* It exposes us to more technical risks -- e.g. (1) if a future Mosaico v0.25 replaces tinymce with something else, then this module gets put in the hard spot of either breaking its contract, bending over backward to translate it, or forking. (2) If we change the baseline UX (e.g. replacing `civicrmtoken` with something else), then we have to work harder to perform/communicate upgrades.
* The option itself is pretty techy. (Then again, "template path" is also pretty techy. I felt a bit uncomfortable with accepting that at the time...)
* There are several more things in the Mosaico config that would be equally interesting to customize (e.g. the token hotlist or the minimal toolbar), and future developers will regard this as a precedent/model.

Let me verbalize a couple ideas for mitigating the risk.

1. Keep the basic design, but -- in the UI -- flag the settings as "EXPERIMENTAL".
2. Change the basic design -- instead of a "Setting", add `hook_civicrm_mosaicoConfig(&$config)`. Add an example in Markdown docs, but flag the hook as "EXPERIMENTAL". (Optionally, this patch could be converted into a new `com.cividesk.mosaicotweaker` extension.)

Mulling this... I'm really not keen on upgrade-maintenance implications of `#1` (setting-workflow) but could acquiesce if a couple more folks said "The settings UX is good." ...  Overall I'm leaning toward `#2` (hook-workflow), which I think will (a) give you *more* ability to customize, (b) set the bar high-enough that any customizers know what trouble they're getting into, and (c) keep the risks out of this module.

(CC @deepak-srivastava @mattwire @MegaphoneJon @jamienovick - interested in second opinions on customization UX)

nganivet:
@totten Changing the icons on the toolbar is a very common customer request, and this is an elegant way to resolve it. Going through a hook over-complicates such trivial change very significantly, and will make it a lot more expensive for the end-users (1-2 hours of work to create/test a custom extension vs a 1 minute change in the UI).

I appreciate that this is a technical setting, but:
- as you rightly pointed out, it is not more technical than having a path settings whose default value is [civicrm.files]/mosaico.tpl
- as @mattwire pointed out we could add contextual help to this field
- and finally, there is very good documentation on these plugins and icons on the tinyMCE website, which we could point to in the contextual help

sunilpawar:
@mattwire @nganivet added reset to default value functionality with help text. you can suggest help text for both field.

deepak-srivastava:
Although we haven't come across such requests so far, there is no harm in keeping it. I would put them in a separate fieldset (collapsed by default) and call it "Advanced / experimental" due to the settings being techy. This way it won't clutter basic user settings.

A few other observations:
The values may need to be sanitized before it goes into database. Or design would need to be tweaked so values are distinct & concrete (type check) - e.g multiselect or something.
I would use api for CRM_Core_BAO_Setting::getItem().

nganivet:
@sunilpawar Can you please make relevant changes to implement these suggestions? Thanks.

veda-consulting:
@sunilpawar @nganivet I don't think this logic should be in the Mosaico extension as its an edge case that isn't part of the mosaico functionality so to speak. I think the compromise is that Mosaico can warn about it if its aware but the cure should be in an specific extension which deals with it. Its not an issue we have seen come up at any site we've installed at and it adds an option that is outside of the Mosaico feature base. I hope that makes sense?

nganivet:
@veda-consulting This PR adds a new setting allowing administrators to change the TinyMCE configuration, and in particular the icons available in the toolbar. So it does not resolve any issues as you above post implies. This feature has been requested by probably 50% or more of my customers using Mosaico, primarily for the following: adding the numbered and/or bulleted list icons, and adding the RTL/LTR icons in order to be able to write emails in RTL languages. I therefore think this feature should be accepted as fulfilling a common customer need that cannot currently be solved otherwise. It also cannot be put in an extension unless you add some hooks in the Mosaico extension itself, and adding those hooks will probably be way more complicated than this simple patch. Thanks.

homotechsual:
I agree that this feature set is needed, it is not outside the realm of what Mosaico itself does - their own site https://mosaico.io implements custom TinyMCE options to allow (for example) lists and Mosaico explicitly supports sending options to TinyMCE though it's own code/implementation.

At the moment the lack of common formatting options like lists severely limits the usefulness of this extension this PR proposes a sensible, if not perfect, method to address this issue in my opinion.

deepak-srivastava:
Just discussed with @veda-consulting. If @sunilpawar can make the changes suggested, think it balances both the sides for now, and we ok to merge the PR.

JKingsnorth:
Wow, well I just created a separate PR before reading this thread, which implements exactly what @totten describes above: https://github.com/veda-consulting/uk.co.vedaconsulting.mosaico/pull/272

I think that the hook_civicrm_mosiacoConfig is definitely the best way to go about this. It offers the best flexibility, and doesn't pollute the admin interface with techy fields that are tricky for users to understand what they are doing.

> adding those hooks will probably be way more complicated than this simple patch.
@nganivet 

Actually, the patch to add the hook is way simpler I think, thanks to the super-simple interface that @totten has built! Check out the linked PR.

And the hook has the benefit of being able to customise other aspects of the editor configuration, for both the limited and 'full' versions of the editor, and changing 'base' settings, like `['forced_root_block'] = FALSE`

deepak-srivastava:
Closing as #272 has been merged in. Thanks everyone.

fulltrucker:
For anyone still interested in a way to configure the Mosaico toolbar in the Civi UI, we created this extension that implements (near as I can tell) a similar approach that was initially outlined in this issue. You can grab it here: https://github.com/ginkgostreet/com.ginkgostreet.mosaicotoolbarconfig

- It uses the hook that was implemented in #272
- It may not have some of the additional features discussed in this issue i.e. contextual help 
- Are these settings a bit more "techy"? Sure... and this has been discussed before and now it's housed in an extension and therefore optional... you can decide if your client is capable of handling it. 
- If any of the work done in this PR that didn't make it to the Mosaico extension would fit in the Configurator extension (again contextual help comes to mind) then by all means fork it and do a PR. We're happy to have the community help improve the extension.

We have successfully installated the extension (after applying the changes in #272) and enabled RTL/LTR buttons in the editor of a client who needs this capability as many of their emails are sent in Hebrew. They also wanted bulletted lists and a couple other whiz-bang options.

PR #256, opened 2018-06-06
CIVICRM-886: Update injected parameters for crmMosaicoBlockMailing

agileware-fj:
#249 introducted a regression where the directive definition crmMosaicoBlockMailing does not have crmMailingSimpleDirective injected, preventing the mailing UI from loading at all in our testing.

This adds it back into the parameters, allowing the interface to load again.


jusfreeman:
Ping @seamuslee001

seamuslee001:
ping @totten Tim i think this overlooked, i think this might be a better solution to what you were solving here https://github.com/veda-consulting/uk.co.vedaconsulting.mosaico/pull/257/files

totten:
It's basically the same solution, but I hadn't noticed this PR at the time, and this one is a bit more tidy. 👍  Merged manually. Thanks, @agileware-fj.

Issue #251, opened 2018-05-02
TinyMCE editor plugin settings


sunilpawar:
Allow admin to add/remove tinymce editor plugin.

domain-name/civicrm/admin/mosaico?reset=1
<img width="721" alt="mosoico_admin" src="https://user-images.githubusercontent.com/377735/39509152-c4c785a4-4e02-11e8-994f-9dbfb2363d6c.png">

sunilpawar:
#250

homotechsual:
We've taken a different approach to this by changing the defaults to those used on the mosaico demo.

mattwire:
Closing following 2.6 release. Please reopen if still an issue

Issue #238, opened 2018-02-12
Text block is too basic - ordered/unordered lists + add image


KairosCanada:
Why isn't there an option to add a bulleted list in a text block?
Why can't we add a photo and wrap text around it?
Why can't we upload photos via URL?

These are some basic things that you will need to develop if you want users to stick with Mosaico.

Staff not familiar with HTML expect templates to work like a Word document. If you ask them to look at the source code they get flustered and won't use it, even for something as simple as bullets.

MegaphoneJon:
@KairosCanada - I'm not one of the devs of this extension, but the short answer is "because no one has written these features".  

A slightly longer answer:
* Mosaico is a project that's separate from CiviCRM.  This repository merely packages Mosaico for use within CiviCRM.  Feature requests for Mosaico itself should go to the [Mosaico team](https://github.com/voidlabs/mosaico).
* While this software is free to use, it isn't free to write.  It's not effective to demand something of people who have already given you so much of their work with no benefit to themselves.  Offering resources to help them meet your goals is a better approach.  If your organization is unwilling/unable to prioritize those goals, your best hope is that someone else with your goals feels differently.

themak1985:
@KairosCanada - @MegaphoneJon  is right - if the dev-team isn't supported to implement the features, they won't be implemented. by support I mean $$$ or development time (coding).

- Bulleted lists is a feature that has been implemented in the latest Mosaico - not sure when exactly it was implemented, but its in v0.17.3 - however - it will need to be implemented into this CiviCRM version. 
- As far as the other features, such as wrapping text around an image - its my understanding you can do this via html or through integration with TinyMCE - though you would again need to provide $$$ or development time to implement this.

tbuytaer:
I would be interested in having the bulleted list button too.

I found this issue from upstream mosaico that says it can be done by adding a "mosaico plugin" but I don't know how this should be set. If someone can give me a pointer I can try to see if I can do this.
https://github.com/voidlabs/mosaico/issues/194

homotechsual:
We've issued a PR to change the defaults for the editor in CiviCRM's implementation of Mosaico. It involves no changes to Mosaico instead overriding these at the point the CiviCRM calls Mosaico.

PR #264, opened 2018-07-09
Change WYSIWYG to match experience offered on https://mosaico.io

homotechsual:
This PR simply changes the default WYSIWYG config to match that offered by https://mosaico.io - adding in some missing options like Bulleted Lists, Numbered Lists.

Linked issues: #238 #251 

**NB:** This PR takes a different approach to #251, instead of making this a configurable option this sets the default implementation in a manner which can still be overridden/changed in a later PR.

### DIFF
--- a/CRM/Mosaico/Page/Editor.php
+++ b/CRM/Mosaico/Page/Editor.php
@@ -96,6 +96,137 @@ class CRM_Mosaico_Page_Editor extends CRM_Core_Page {
       'tinymceConfigFull' => array(
         'plugins' => array('link hr paste lists textcolor code civicrmtoken'),
         'toolbar1' => 'bold italic forecolor backcolor hr styleselect removeformat | civicrmtoken | link unlink | pastetext code',
+        'style_formats' => array(
+          0 => array(
+            'title' => 'Headings',
+            'items' => array(
+              0 => array(
+                'title' => 'Heading 1',
+                'format' => 'h1',
+              ),
+              1 => array(
+                'title' => 'Heading 2',
+                'format' => 'h2',
+              ),
+              2 => array(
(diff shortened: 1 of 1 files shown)


homotechsual:
This is achieved the same way Mosaico themselves achieve it - by defining the TinyMCE formats menu as an array. The array in question is mostly cribbed directly from their JS, then converted from a JSON array to a PHP array, tidied up and tested.

mattwire:
@MikeyMJCO I just had this requirement and applied your patch  - seems to work well.  Can you point to where you found:
> the default WYSIWYG config to match that offered by https://mosaico.io

In other words, is there code here: https://github.com/voidlabs/mosaico that uses these defaults?

homotechsual:
Hi @mattwire when I say defaults I mean on their implementation at https://mosaico.io - which is where the code in question comes from. The default GitHub version doesn't include the changes as they say that it's configurable by implementers.

mattwire:
@MikeyMJCO Looking at #251 it seems that they are adding lists simply by changing the default for toolbar1.  What is the difference between doing that and adding via style_formats here?

mattwire:
@sunilpawar Any thoughts on this approach? Instead of what you proposed in #251

JKingsnorth:
I propose this approach instead:
https://github.com/veda-consulting/uk.co.vedaconsulting.mosaico/pull/272

Adding a custom hook will allow different sites to set their own configuration options, which best suit their requirements. I think this is the most flexible and maintainable approach. Keep a 'limited' amount available by default, and allow sites to add to it if they wish.

homotechsual:
I'm closing this in favour of what @JKingsnorth proposes - much more elegant!

#272 is the right way to implement this in the long run IMO.

PR #271, opened 2018-08-17
add domain change edits

Stoob:
added query to handle domain switch


mattwire:
@totten @deepak-srivastava Could this please be merged. It's simple docs improvements only

PR #272, opened 2018-08-26
Add a hook for Mosaico editor configuration

JKingsnorth:
We'd like to customise the WYSIWYG editor settings, specifically to add the tinymce config setting:
`$config['tinymceConfig']['forced_root_block'] = FALSE;`

Rather than overriding the code, it would be great to be able to customise the configuration through a hook interface.

This PR adds a hook_civicrm_mosaicoConfig - named after the function that it can modify the output of.

### DIFF
--- a/CRM/Mosaico/Page/Editor.php
+++ b/CRM/Mosaico/Page/Editor.php
@@ -61,7 +61,7 @@ class CRM_Mosaico_Page_Editor extends CRM_Core_Page {
       'sequential' => 1,
     ));
 
-    return array(
+    $config = array(
       'imgProcessorBackend' => $this->getUrl('civicrm/mosaico/img', NULL, TRUE),
       'emailProcessorBackend' => 'unused-emailProcessorBackend',
       'titleToken' => 'MOSAICO Responsive Email Designer',
@@ -98,6 +98,17 @@ class CRM_Mosaico_Page_Editor extends CRM_Core_Page {
         'toolbar1' => 'bold italic forecolor backcolor hr styleselect removeformat | civicrmtoken | link unlink | pastetext code',
       ),
     );
+
+    // Allow configuration to be modified by a hook
+    if (class_exists('\Civi\Core\Event\GenericHookEvent')) {
+      \Civi::dispatcher()->dispatch('hook_civicrm_mosaicoConfig',
+        \Civi\Core\Event\GenericHookEvent::create(array(
+          'config' => &$config,
+        ))
+      );
+    }
+
+    return $config;
   }
 
   /**


JKingsnorth:
Actually, what would be even better would be to be able to customise the configuration per template. So I'm going to work on this to allow that, provide some rationale, and add the hook documentation and examples to the api readme as well.

sunilpawar:
@JKingsnorth  i tested hook_civicrm_mosaicoConfig functionality it worked as expected. We can customize toolbar , plugin using this hook.

JKingsnorth:
Great! I think this is the best solution to these issues, and I left some more explanation/reasoning in: https://github.com/veda-consulting/uk.co.vedaconsulting.mosaico/pull/250#issuecomment-416074776

homotechsual:
I've closed #264 as I think this is a much better approach!

mattwire:
Right, so this looks like the way to go.  Do any of you have an extension you've written to take advantage of this hook so we can test it easily?

JKingsnorth:
I have implemented the hook in some custom code. If it needs a test then I'm happy to take advice on the best approach to take. I also think I should add something to the documentation!

deepak-srivastava:
Would be good to have a documentation for implementation example as well. 
Thanks to @JKingsnorth, @sunilpawar, @mattwire, @MikeyMJCO :+1:

JKingsnorth:
Thanks for the merge @deepak-srivastava

fulltrucker:
> Would be good to have a documentation for implementation example as well.
> Thanks to @JKingsnorth, @sunilpawar, @mattwire, @MikeyMJCO 👍

+1 to both comments above.

We have a client asking about customizing their editor, not sure how to use this hook. I assume via a custom extension? Could said extension theoretically have a WYSIWYG-style configurator (sorta like /civicrm/admin/ckeditor?reset=1) that uses this hook to customize the editor?

JKingsnorth:
I will try to write the documentation tonight.

In theory it would be possible to write an extension which uses this hook to create a drag and drop style configuration page.

However, as suggested above, it feels more likely that the mosaico WYSIWYG would be configured once when a developer implements a new template.

JKingsnorth:
I've created a PR for the hook's documentation here:
https://github.com/veda-consulting/uk.co.vedaconsulting.mosaico/pull/289

fulltrucker:
For anyone interested in a way to configure the Mosaico toolbar in the Civi UI, we created this extension that does that: https://github.com/ginkgostreet/com.ginkgostreet.mosaicotoolbarconfig

It's not drag and drop, just a couple of text fields, but it gets the job done and works like a champ so far as we can tell. Thanks to everyone who worked on this hook.

PR #276, opened 2018-11-05
Update Mosaico admin form for 5.8

eileenmcnaughton:
Without this update Mosaico will be using a deprecated format for
declaring html_type once it hits 5.8 and there will be
deprecation notices (only visible on dev sites).

But this PR goes further & removes the entire form in favour of
a generic approach per https://github.com/civicrm/civicrm-core/pull/13054

(don't merge this until that is merged)

also wait for this to be merged https://github.com/civicrm/civicrm-dev-docs/pull/555 and follow on updates to docs.

Also note - I incremented the xml file to 5.8 - that is the methodology I use - basically
I regularly increment the version  (especially if there is change) so that sites on
the later version get the updated version but there is still a version for the older sites that
works on theirs - I realise in the internal beta-land of Mosaico that might not quite work :-)

However, the minimum required to extend the settings form (which is officially not
a supported integration point) and not get deprecation notices is to alter the case
from 'Text' to 'text' when you declare html_type.


totten:
I've rebased this -- splitting into separate commits for (a) the minimal part and (b) the maximal part. The minimal part is now merged.

Issue #278, opened 2018-11-09
Templates do not load in New Mailing (Traditional) 


John-B:
after installing Mosaico, calling MessageTemplate get in the API explorer returns and empty array; and the dropdown for selecting a Message Template which appears when visiting CiviCRM > Mailings > New Mailing (Traiditiona) is not populated.

After installing Mosaico a schema class is added by the extension's schema  to civicrm/CRM/Core/DAO/AllCoreTables.data.php for the table civicrm_msg_template. Its name is MessageTemplate. This has the same name as the class defined in  civicrm-core/xml/schema/Core/MessageTemplate.xml 
For reasons I do not understand this clash of naming creates a problem, and I have discovered that manually hacking civicrm/CRM/Core/DAO/AllCoreTables.data.php to change 
`  'CRM_Mosaico_DAO_MessageTemplate' => [
    'name' => 'MessageTemplate',
    'class' => 'CRM_Mosaico_DAO_MessageTemplate',
    'table' => 'civicrm_mosaico_msg_template',
  ],
`
to 
`  'CRM_Mosaico_DAO_MessageTemplate' => [
    'name' => 'MosaicoMessageTemplate',
    'class' => 'CRM_Mosaico_DAO_MessageTemplate',
    'table' => 'civicrm_mosaico_msg_template',
  ],
`
to remove the name clash with the core MessagesTemplate fixes the problem.

bazzspeed:
This does not appear to be an issue with my setup, can you please retest and also comment on what versions you're using.

_My setup:_
    
    CiviCRM: 5.11 and all extensions are up-to-date:
    Related Permissions Module (nz.co.fuzion.relatedpermissions) version 1.6
    FlexMailer (org.civicrm.flexmailer) version 1.0-beta1
    CiviCRM Multisite (org.civicrm.multisite) version 2.6
    CiviCRM Bootstrap theme (org.civicrm.shoreditch) version 0.1-alpha30
    Mosaico (uk.co.vedaconsulting.mosaico) version 2.0-beta4.1544700816

MegaphoneJon:
Same as @bazzspeed here - same version of Civi, all extensions up to date, running the master branch of Mosaico.

@John-B has this issue been fixed?  If so we can close this.

John-B:
[...]

PR #279, opened 2018-11-12
change name to prevent name clash

John-B:
'MessageTemplate' has the same name as a class in core, and the clash prevents  API calls from working correctly. See issue #278

### DIFF
--- a/xml/schema/CRM/Mosaico/MessageTemplate.xml
+++ b/xml/schema/CRM/Mosaico/MessageTemplate.xml
@@ -2,7 +2,7 @@
 
 <table>
   <base>CRM/Mosaico</base>
-  <class>MessageTemplate</class>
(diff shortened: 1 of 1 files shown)


mattwire:
This is the right fix, however it's slightly complicated as the schema is used to generate the DAO so the CRM_Mosaico_DAO_MessageTemplate class needs renaming too and checking if it's in use anywhere.

bazzspeed:
Is this issue a blocker for stable release of 2.0 ?

John-B:
I think it is a blocker because it means there is a link for creating a traditional (i.e. non-Mosaico) email in the menu, but once Mosaico is installed that link does not work; and some people who install Mosaico may want to keep the possibility of creating a 'traditional' email.

mattwire:
Now resolved see #278

PR #280, opened 2018-11-16
Replacing 'TITLE' by mail subject in template HTML

mfrausto-mdw:
In the mailings sent from CiviCRM, there's usually a "View in your browser" link (for when the mailing renders badly in the user's email client). When you visit that page, the page title is simple "TITLE".

This fix is replacing 'TITLE' by e-mail subject in generated HTML

This has been previously reported:
https://github.com/veda-consulting/uk.co.vedaconsulting.mosaico/issues/245

### DIFF
--- a/ang/crmMosaico/MixinCtrl.js
+++ b/ang/crmMosaico/MixinCtrl.js
@@ -57,6 +57,12 @@
           // Mosaico exports JSON. Keep their original encoding... or else the loader throws an error.
           mailing.template_options.mosaicoMetadata = viewModel.exportMetadata();
           mailing.template_options.mosaicoContent = viewModel.exportJSON();
+          /*
+          Fix to display mailing subject instead of "TITLE"
+          Note that we're not validating subject since is a required field
+          and there is only one "TITLE" in the template
+           */
+          mailing.body_html = mailing.body_html.replace("TITLE", mailing.subject);
         }
 
         crmMosaicoIframe = new CrmMosaicoIframe({


totten:
Interesting find/bug. Thank you for the first time contribution here!

The patch as written gives me a somewhat queasy feeling:
* It's not hard to organically have the text `TITLE` floating around different parts of the message. I suspect the preconditions *right now, in typical usage* allow this to work, but it's really not clear how strongly the preconditions are enforced.
* The `mailing.subject` is edited in a different way from the main content -- which means that they change at different times. (Example: *Create a mailing with a draft title; then design some Mosaico content; then go back and edit the subject. Does the title in the Mosaico content have the old or new subject?* There are probably other example flows based on resuming drafts or reusing templates.)

Ideas/suggestions:
* Use a stronger matching criterion. If you can't find a way to rely on a real HTML parser, then at least match on `;<title>\s*title\s*</title>;i`or somesuch.
* Instead of putting in the literal subject at the time of saving, use a mail-merge token like `[subject]` or `{mailing.subject}` -- and let the server-side handle the substitution at the standard substitution time(s).

dsnopek:
Re:

> * use a mail-merge token like `[subject]` or `{mailing.subject}`

Are one of those Mosaico-style (as opposed to CiviCRM-style) tokens? If so, maybe this could be fixed in the template (versafix) itself by changing it to use a Mosaico-style token instead of the text 'TITLE'?

totten:
>  If so, maybe this could be fixed in the template (versafix) itself by changing it to use a Mosaico-style token instead of the text 'TITLE'?

Yeah, editing the template is just as good IMHO as filtering on `<title>TITLE</title>`. (More correct logic - albeit slightly more patch-juggling.) FWIW, the templates in this extension are read from this branch: https://github.com/civicrm/mosaico/tree/v0.15-civicrm-2.

> Are one of those Mosaico-style (as opposed to CiviCRM-style) tokens?

`[subject]`... it might be. I haven't been able to find a canonical list of Mosaico-style mail-merge tokens - the best lists I can see:

* `versafix-1` uses a few tokens (*currently `[unsubscribe_link]`, `[show_link]`, `[profile_link]`, `[mail]`*)
* [mosiaco_civicrm_alterMailContent()](https://github.com/veda-consulting/uk.co.vedaconsulting.mosaico/blob/806f2e709d3b7caaef1e0c5bed43dc21977fd813/mosaico.php#L341-L346) translates a few tokens (*currently `[unsubscribe_link]`, `[show_link]`, `[subject]`*).

There's some impedance mismatch there -- ie `alterMailContent` supports `[subject]` even though upstream `versafix-1` doesn't use it. (Tangentially: https://github.com/voidlabs/mosaico/issues/488)

mfrausto-mdw:
Will be closed in favour of:
https://github.com/civicrm/mosaico/pull/4

rameshrr99:
Still getting the same error message - I am using latest release. Also I can see all the patch are up to date

Also when I click on "view on browser" link _mosaico_civicrm_alterMailContent function is not called at all

PR #286, opened 2019-01-03
Mark Mosaico 2 stable

eileenmcnaughton:
@veda-consulting @deepak-srivastava @totten mattermost discussions amounted to 'there are no more blockers' - shall we do it!

### DIFF
--- a/info.xml
+++ b/info.xml
@@ -15,8 +15,8 @@
     <url desc="Licensing">http://www.gnu.org/licenses/agpl-3.0.html</url>
   </urls>
   <releaseDate>2017-12-07-</releaseDate>
-  <version>2.0-beta4</version>
-  <develStage>beta</develStage>
+  <version>2.0</version>
+  <develStage>stable</develStage>
   <compatibility>
     <ver>5.0</ver>
   </compatibility>


francescbassas:
Maybe it is still necessary to go through a beta5 because there is no release since jun 2018 and there are significant changes like https://github.com/veda-consulting/uk.co.vedaconsulting.mosaico/pull/281.

MarshCastle:
Just through curiosity - can this be marked as stable if the dependencies aren't yet stable?

eileenmcnaughton:
I believe that's circular - @totten told me flex mailer isn't stable because mosaico isn't

mattwire:
Let's just mark it stable! Dependencies can follow. Date in the info.xml will need updating too...

francescbassas:
> Just installed Mosaico 1.1 for CiviCRM 5.7.2 on Drupal 7.

See https://civicrm.stackexchange.com/q/28466/104

mattwire (changes requested):
Note the release date needs updating before release...

mattwire:
@totten @deepak-srivastava @veda-consulting Any progress on this - can we mark what we've got as 2.0 stable?

We're going to need to increment min version soon to support things like this: https://github.com/veda-consulting/uk.co.vedaconsulting.mosaico/pull/299 so now seems like a good time...

mattwire:
Merged via #300

PR #298, opened 2019-03-22
Allow sending traditional ad-hoc emails as a search task

MegaphoneJon:
Newer betas of Mosaico 2.x allow creating a non-Mosaico CiviMail through **Mailings » New Mailing (Traditional)**.  However, there is no equivalent for creating an ad-hoc mailing via the search action "Email - Schedule/send via CiviMail".

This PR adds a new search action: "Email - Schedule/send via CiviMail (traditional)" which allows creating ad-hoc mailings using traditional templates.

### DIFF
--- a/CRM/Mosaico/Form/Task/AdhocMailingTraditional.php
+++ b/CRM/Mosaico/Form/Task/AdhocMailingTraditional.php
@@ -1,0 +1,40 @@
+<?php
+
+/**
+ * Given the selected contacts, prepare a *traditional* mailing with a hidden group.
+ */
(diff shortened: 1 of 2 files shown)


mattwire:
This feels like it fixes a bit of missing functionality.  However, I'm wondering if we should also add a mosaico setting at the same time that enables/disables the "traditional mailing" option altogether - as some sites only want to use mosaico, while others may want to use both.

Thoughts on merging @totten @eileenmcnaughton ?

mattwire:
@MegaphoneJon Thanks

PR #299, opened 2019-04-26
Always place iframe below KAM menu instead of underneath.

mattwire:
Since the introduction of the KAM menu into core we have a consistent css ID across CMSs `#civicrm-menu`.

So we can: 
1. Simplify code to position the iframe.
2. Always align the iframe below the KAM menu no matter what position it is currently in (ie. overlapping/below the CMS menu).

I've tested this on Joomla, Drupal7 and Wordpress and seems to work well on all three.  Only "nice-to-have" is to trigger a reposition of the iframe if the KAM menu is moved (as is triggered on window resize) but that doesn't feel essential and I couldn't see how to trigger that easily.

Ping @colemanw

### DIFF
--- a/ang/crmMosaico/Iframe.js
+++ b/ang/crmMosaico/Iframe.js
@@ -26,15 +26,7 @@
           var c = CRM.crmMosaico || {};
           var top = 0, left = 0, width = $(window).width(), height = $(window).height();
           if (c.topNav && $(c.topNav).length > 0) {
-            if (c.drupalNav && $(c.drupalNav).length > 0 && $(c.drupalNav).outerHeight() > $(c.topNav).outerHeight()) {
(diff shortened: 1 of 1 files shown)


colemanw (approved):
This looks right to me.

totten:
> I've tested this on Joomla, Drupal7 and Wordpress and seems to work well on all three.

🎉 👍 🎉

@mattwire Since this specifically targets KAM, should we update `info.xml` and bump `<ver>` to `5.12`?

mattwire:
> @mattwire Since this specifically targets KAM, should we update `info.xml` and bump `<ver>` to `5.12`?

Yes, though let's bump ver to 5.13 to align with next release requirements of flexmailer

PR #315, opened 2019-06-25
Drop check for shoreditch as there is no loss in user experience, just a slightly different look and feel

mattwire:
And users may prefer to use alternative theme extensions (eg. https://github.com/mattwire/civicrm-haystacktheme)


eileenmcnaughton:
This has a couple of up-votes so seems non-controversial - merging

PR #322, opened 2019-07-24
Remove legacy template field

agileware-pengyi:
The legacy mailing template is not used in Mosaico template, however, the field is shown when editing Moscaico mailing **without** shoreditch. This cause users to confuse.

Before
------
Legacy template field in the form.

After
------
Legacy template field removed from the form.

Comment
------
Feel free to improve this PR since I am not familiar with Angular.

Agileware ref: CIVIMOSAIC-18

### DIFF
--- a/ang/crmMosaico.crmstar/EditMailingCtrl/crmstar-single.html
+++ b/ang/crmMosaico.crmstar/EditMailingCtrl/crmstar-single.html
@@ -6,7 +6,90 @@
           <div crm-ui-tab-set>
             <div crm-ui-tab id="tab-mailing" crm-title="ts('Mailing')">
               <div crm-mailing-block-summary crm-mailing="mailing"></div>
-              <div crm-mailing-block-mailing crm-mailing="mailing"></div>
+              <div>
+                <div class="crm-block" ng-form="subform" crm-ui-id-scope>
+                  <div class="crm-group">
+                    <div crm-ui-field="{name: 'subform.fromAddress', title: ts('From'), help: hs('from_email')}">
+                      <div ng-controller="EmailAddrCtrl" crm-mailing-from-address="fromPlaceholder"
+                           crm-mailing="mailing">
+                        <select
+                                crm-ui-id="subform.fromAddress"
+                                crm-ui-select="{dropdownAutoWidth : true, allowClear: false, placeholder: ts('Email address')}"
+                                name="fromAddress"
+                                ng-model="fromPlaceholder.label"
+                                required>
+                          <option value=""></option>
+                          <option ng-repeat="frm in crmFromAddresses.getAll() | filter:{is_active:1} | orderBy:'weight'"
+                                  value="{{frm.label}}">{{frm.label}}
+                          </option>
+                        </select>
+                      </div>
+                    </div>
+                    <div crm-ui-field="{name: 'subform.replyTo', title: ts('Reply-To')}"
+                         ng-show="crmMailingConst.enableReplyTo">
+                      <div ng-controller="EmailAddrCtrl">
+                        <select
+                                crm-ui-id="subform.replyTo"
+                                crm-ui-select="{dropdownAutoWidth : true, allowClear: true, placeholder: ts('Email address')}"
+                                name="replyTo"
+                                ng-change="checkReplyToChange(mailing)"
+                                ng-model="mailing.replyto_email"
+                        >
+                          <option value=""></option>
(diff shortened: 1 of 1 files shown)


eileenmcnaughton:
@agileware-pengyi just very superficially - the description makes me think you would be REMOVING code but you are adding a lot - which makes me wonder why... Perhaps screenshots would help

agileware-pengyi:
@eileenmcnaughton the removed line reusing [this file](https://github.com/civicrm/civicrm-core/blob/d161ee89826cb74dda8c7b47bf25d7891176b54f/ang/crmMailing/BlockMailing.html#L9). I replaced the line with its content and remove the template field. So the field will not be displayed in Mosaico editor.
mailing_block

seamuslee001:
ok so looking at this the concept of the PR makes 100% sense to me, the way the PR has been done doesn't 100% make as much sense to me but that is ok. Just verballing out there are possibly other options we could pursuse

1. Alter CiviCRM-Core's BlockMailing.html to do an ng-if on the template if the type is not traditional
2. We could create our own new directive here and specifiy it like the block Mailing, This would potentially have the advantage of having a structure that is similar to the base stuff so developers may understand things easier
3. Do as this PR is doing which is bringing the code we want from the BlockMailing directive into the main file here

@totten @eileenmcnaughton @agileware-pengyi i'm not sure what is overall the best option here but my thinking would be 1 would be the most simple to do and possibly have the best results overall for any future template builder (may not necessairly be Mosacio).

eileenmcnaughton:
1 seems Ok to me - but I'm less familiar with this code

agileware-pengyi:
I would prefer 1 if the traditional templates won't be used on any other editors.

seamuslee001:
I would assume that traditional templates wouldn't be used at all going forward for newer editiors. I would appreciate @totten weighing in as well

totten:
Agree, it's really nice how `#1` avoids duplicating the code. The essential logic also sounds valid: "content from `civicrm_msg_template` can be used if-and-only-if the mailing has `template_type==traditional` (*because the GUI for editing `civicrm_msg_template` is geared toward `traditional`*)"

In the long-term, it might be ideal for `civicrm_msg_template` to also have a `template_type` (and bunch of other changes to make that workable). However, in the foreseeable future, I think `#1` is a straight-forward improvement.

totten:
> Feel free to improve this PR since I am not familiar with Angular.

@agileware-pengyi - Just want to say - for someone coming in new to Angular, it looks like a good start. 👍 

I know more feedback here becomes moot when changing to `#1`, but it may be useful to follow-through with a little feedback anyway. Consider these files:

* `mosaico` - ` ang/crmMosaico.crmstar/EditMailingCtrl/crmstar-single.html`
* `mosaico` - `ang/crmMosaico.bootstrap/EditMailingCtrl/bootstrap-single.html`
* `mosaico` - `ang/crmMosaico.bootstrap/EditMailingCtrl/bootstrap-wizard.html`
* `core` - `ang/crmMailing/EditMailingCtrl/2step.html`
* `core` - `ang/crmMailing/EditMailingCtrl/unified.html`
* `core` - `ang/crmMailing/EditMailingCtrl/unified2.html`
* `core` - `ang/crmMailing/EditMailingCtrl/wizard.html`
* `core` - `ang/crmMailing/EditMailingCtrl/workflow.html`

All of these files are *basically* the same, except that they remix the various UI elements in different layouts. The current set of directives (`crm-mailing-block-summary`, `crm-mailing-block-mailing`, `crm-mailing-block-preview`, etc) was meant to facilitate this.

The change here (hiding one item from `crm-mailing-block-mailing` in the context of Mosaico) is not easy with the current set of blocks/options - they just don't support that nicely. The approach in 322 (i.e. inlining the block completely) is understandable but feels a bit heavy-handed. Both Seamus's `#1` and `#2` feel better than `#3` because they still support some remixability.

There is a good argument that maybe the contract for these directives (`crm-mailing-block-summary` etc) should be made for flexible. I don't see a *perfect alternative contract for all possible scenarios* off-the-cuff, but this seems like an example of a better one: `crm-mailing-ab-block-mailing="{field:bool, field:bool, ...}"`)

https://github.com/civicrm/civicrm-core/blob/7cded6f/ang/crmMailingAB/EditCtrl/edit.html#L33-L51

That allows all of the wonky field-markup to be defined one time (`crmMosaicoAB/BlockMailing.html`) while also allowing elements to be rearranged/cherry-picked.

jusfreeman:
Just noting that alternative PR has been submitted here https://github.com/civicrm/civicrm-core/pull/14927

PR #323, opened 2019-08-05
Replace the image source to static url and remove placeholder before sending

agileware-pengyi:
Overview
----------
One of our clients is experiencing a performance issue every time when the Mosaico mailing send out. We found that a lot of Mosaico image request is logged and want to reduce the load on each image request.
The URL in Mosaico editor includes an image processor to resize the source image. Though it will cache the resized image, it still needs to bootstrap the CMS and CiviCRM every time.

Before
---------
The image URL is `src="https://domain/civicrm?page=CiviCRM&amp;q=civicrm%2Fmosaico%2Fimg&amp;src={source_url}&amp;method=resize&amp;params=258%2Cnull"`

After
---------
The image URL is `https://domain/wp-content/uploads/civicrm/persist/contribute/images/uploads/static/{filename}`

Note
---------
I tried to find a way to let Mosaico generating the static URL instead of replacing but cannot find a way to do it.

Agileware ref: CIVIMOSAIC-19

### DIFF
--- a/CRM/Mosaico/Utils.php
+++ b/CRM/Mosaico/Utils.php
@@ -313,6 +313,34 @@ class CRM_Mosaico_Utils {
   }
 
   /**
+   * Called in hook_civicrm_alterMailContent
+   * Generate the static image and return the static url for the image
+   *
+   * @param array $matches
+   *                      The regex matches
+   *
+   * @return string
+   *               The static url
+   */
+  public static function cacheImage($matches) {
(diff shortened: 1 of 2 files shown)


agileware-pengyi:
I will close this PR for more testing. Will reopen it when it is ready.

agileware-justin:
Sounds related to https://github.com/veda-consulting-company/uk.co.vedaconsulting.mosaico/pull/568

PR #327, opened 2019-08-29
Improve browser incompatibility message

aydun:
See https://github.com/veda-consulting/uk.co.vedaconsulting.mosaico/issues/296


mattwire:
Yay for helpful error messages!

Issue #340, opened 2019-10-18
Looking for missing placeholder and social media images?


Stoob:
After a recent upgrade, Mosaico is showing broken 'placeholder' images for upload in the template and returning errors in the Drupal log such as these:

>https://site.edu/civicrm/mosaico/img?method=placeholder&params=166%2C130

This results in the error seen here.
![errors](https://user-images.githubusercontent.com/1683473/67123467-3efa4e00-f1a5-11e9-9bb2-dac5a6b0b50b.png)

While other 'package' images seem to be 404 missing entirely.

>The requested URL "/sites/default/files/civicrm/ext/uk.co.vedaconsulting.mosaico/packages/mosaico/img/social_def/google+_bw_ok.png" was not found on this server.

>https://site.edu/sites/default/files/civicrm/ext/uk.co.vedaconsulting.mosaico/packages/mosaico/img/social_def/google+_bw_ok.png

Mosiaco is otherwise working.  I am able to upload images of my own, but these placeholder and social media files it is looking for appear to be missing entirely.  Indeed there is no 'img' directory within the package at all, neither in the CiviCRM Mosaico extension nor the Mosaico direct distribution itself: https://github.com/voidlabs/mosaico

I am confused by this.  Why is Mosaico looking for images that no longer exist?  @deepak-srivastava Thanks!

mattwire:
The actual `img` directory is under `packages/mosaico/templates/versafix-1/img`. Some earlier versions had issues with image paths and various solutions were put in place on individual sites including symlinks to get around those issues.  However, those issues were fixed before the stable 2.0 came out. But it could be possible that your templates still have hardcoded paths if they've not been updated (there are versions in the db if you don't go back to the base template).

themak1985:
After encountering this issue after upgrading to 2.2 from 2.1 using the CiviCRM UI on D7 we uninstalled mosaico completely and did a fresh install of 2.2 from the CiviCRM extensions directory. The issue persisted. We uninstalled again and installed 2.1 - all issues were resolved.

Thinking it could be an issue with the upgrade that took place, I upgraded to 2.2 via the UI again - issue returned.  We are staying with 2.1 for now.

Lastly - I created a demo wordpress civicrm site on the same server and installed 2.2 via civicrm ui - there were no issues.

aydun:
[...]

PR #342, opened 2019-10-24
Change image placeholder permissions

homotechsual:
Changes the image placeholder permissions check to use OR instead of AND when evaluating required perms.


homotechsual:
Tested this change locally - resolves the issue as presenting in #340 - no ill effects observed - no degradation of security - placeholder generation is correctly restricted to one (or more) of the three permissions.

Issue #312, opened 2019-06-07
Error Loading Images using GD


ryanlrobinson:
The issue occurs when attempting to upload an image in the template builder. We receive the message "Unexpected upload error (Service unavailable (with message))". It also writes to the Drupal log:

Error: Class 'Imagick' not found in CRM_Mosaico_Utils::processUpload() (line 231 of /home/pillar/public_html/sites/default/files/extensions/uk.co.vedaconsulting.mosaico/CRM/Mosaico/Utils.php).

We traced that to find that the code seems to have a clear reference to the Imagick class even though the setting in /civicrm/admin/mosaico?reset=1 is set to "Intervention Image API (gd)" and GD is installed and working. But this code doesn't seem to allow for GD instead of ImageMagick. Please add this support ASAP.

tapashdatta:
Hi @ryanlrobinson Did you find any solution for this issue? thanks

ryanlrobinson:
It's not a true solution, just a workaround, but here's a diff of what my colleague came up with:

```
@@ -228,14 +228,8 @@
           if (move_uploaded_file($tmp_name, $file_path) === TRUE) {
             $size = filesize($file_path);

-            $image = new Imagick($file_path);
-
-            $image->resizeImage($config['THUMBNAIL_WIDTH'], $config['THUMBNAIL_HEIGHT'], Imagick::FILTER_LANCZOS, 1.0, TRUE);
-            // $image->writeImage( $config['BASE_DIR'] . $config[ THUMBNAILS_DIR ] . $file_name );
-            if ($f = fopen($config['BASE_DIR'] . $config['THUMBNAILS_DIR'] . $file_name, "w")) {
-              $image->writeImageFile($f);
-            }
-            $image->destroy();
+            // Imagemagick not working, so simply copy same file as thumbnail
+            copy($file_path, $f);
```

tapashdatta:
@ryanlrobinson Thanks that worked. although not best solution but good workaround.

francescbassas:
Same issue. I suspect it should be fixed in some way similar to

`CRM/Mosaico/Utils.php`
```
-            $image = new Imagick($file_path);
-
-            $image->resizeImage($config['THUMBNAIL_WIDTH'], $config['THUMBNAIL_HEIGHT'], Imagick::FILTER_LANCZOS, 1.0, TRUE);
[...]

PR #345, opened 2019-11-04
Use mosaico graphics service to resize thumbnail.

agileware-fj:
Fix for `Error: Class 'Imagick' not found` when uploading images with only GD extension, per #312

### DIFF
--- a/CRM/Mosaico/Utils.php
+++ b/CRM/Mosaico/Utils.php
@@ -235,14 +235,8 @@ class CRM_Mosaico_Utils {
           if (move_uploaded_file($tmp_name, $file_path) === TRUE) {
             $size = filesize($file_path);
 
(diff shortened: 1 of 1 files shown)


totten:
Agree this code looks pretty sensible. Tested locally with a couple different drivers active, and it seemed to work in each case. There are multiple reports of people using it successfully in #312.

Thank you @agileware-fj @francescbassas !

JKingsnorth:
I think this patch warrants a new minor release, as without it you still need ImageMagick installed alongside GD to make the extension work @totten / @mattwire 

Thanks for the fix!

mattwire:
@JKingsnorth I'd like to get #371 merged if possible before doing another release - be great if you were able to test?

Issue #259, opened 2018-06-08
Spell Checker in Mosaico


kngs:
I have yet to get the EXPERIMENTAL Mosaico Message Templates extension to work to allow spell checking on Mosaico templates (and am also concerned about it's use by the average user as noted in the extension).  Is there any plan and/or timing on getting spell checking directly available in Mosaico?
Thanks - love Mosaico!!

pbatroff:
This extension could help you out: https://github.com/ginkgostreet/com.ginkgostreet.mosaicotoolbarconfig/blob/master/mosaicotoolbarconfig.php

This is what I did to add browser spellcheck to TinyMCE: 

https://github.com/systopia/com.ginkgostreet.mosaicotoolbarconfig/commit/23695d53f9bec8bf86dad09b621a11c8c9eab06f

Good luck!

mattwire:
Fixed in next release after 2.3

PR #352, opened 2019-12-06
Make spell check available directly in mosaico

deepak-srivastava:
While many browsers natively support spell check, TinyMCE has it turned off in mosaico. 
We've had many requests for it, and it really make sense for a template editor to support it by default.

Although it could also be achieved by using extensions that implement mosaicoConfig hook, having another extension just for spell check is maintenance overhead. 

Related issue #259.


mattwire:
I don't see any downside to this - it seems silly to have to add a custom plugin and hook to get a spellchecker - if nobody has any objections I'll merge this?

PR #359, opened 2020-02-10
RSE-777: Limit form changes to Mosaico pages.

reneolivo:
## Description of the problem
CSS styles defined for Mosaico are breaking the styles of other extensions.

The problem is caused because:

1 - form rules are defined to affect any elements in any civicrm page instead of mosaico pages exclusively.
2 - mosaico tries to override shoreditch elements instead of relying on shoreditch to apply these styles.

### Other extension
pr1-before
Notice the remove icon for the "Award Type" and also the smaller width for the date fields.

### Mosaico Mailing Configuration
pr2-before
Notice how the down arrow is lower than it should be, and the remove icon for the "Recipients" fields overlaps with the container's border.

### Other extension
img-after

### Mosaico Mailing Configuration
img2-after

## Technical details

There are several issues, but the main problem is that the styles defined in `sass/civicrm-override/_form.scss` are targeting all form elements in CiviCRM pages instead of exclusively mosaico pages (`.crm-mosaico-page`). To fix this we wrapped all the rules of this file inside a root `.crm-mosaico-page` selector.

In order to compile the CSS we needed to get Mosaico up to date with the latest shoreditch:

- Removed a reference to a `ui-dialog.scss` file, because this file was removed from Shoreditch. It's important to note that shoreditch adds styles to dialogs automatically and there is no need to manually include the styles for them or for any other component so long as the Shoreditch extension is active.
- Updated the variable names for some colours.
- Removed rules to override the styles for Select2 elements. These rules are already being overridden by Shoreditch and there's no further need to override them. In fact, they are clashing since the rules in Mosaico are outdated. This override file was first introduced in https://github.com/compucorp/uk.co.vedaconsulting.mosaico/pull/9 and it's not clear the reason to introduce it. What should be clear is that global style overrides should reside in Shoreditch and not in a particular extension as to avoid clashes.
- Removed some unnecessary rules for help icons since these rules are already present in shoreditch.

## Tests

Backstop tests ran successfully. The only changes are the ones expected. For example:

### Before
backstop_mosaico_Mosaico_-_Mailing_list_-_Advance_Mailing_Options_-_Responses_0_ui-dialog_0_desktop

### After
backstop_mosaico_Mosaico_-_Mailing_list_-_Advance_Mailing_Options_-_Responses_0_ui-dialog_0_desktop (1)


mattwire:
@reneolivo Can you add screenshots for all three display modes please - single page (crm-*), single page (bootstrap) and wizard (bootstrap)?

reneolivo:
@mattwire @totten here are the screenshots:

### Wizard (Bootstrap)
Step 1
Screen Shot 2020-02-18 at 7 23 32 PM
Step 2
fsadfg
Design editor
Screen Shot 2020-02-18 at 7 27 40 PM
Step 3
Screen Shot 2020-02-18 at 7 25 47 PM
Note that the weird radio button spacing comes from Shoreditch and not from this change.

### Single Page Bootstrap
traditional-bootstrap-1
traditional-bootstrap-2
traditional-bootstrap-3
traditional-bootstrap-4
traditional-bootstrap-5
traditional-bootstrap-6
traditional-bootstrap-7

### CRM single page

crm-1

(Only adding one screen because no changes were made to the crm CSS file (`css/mosaico-crmstar.css`)

reneolivo:
@mattwire @totten We also did some investigation and the issue is caused by deploying mosaico in the general CiviCRM angular path instead of a custom path just for Mosaico. ie. current:`/civicrm/a/` desired: `/civicrm/a/mosaico/`

By default all angular modules are bootstrapped into the `/civicrm/a/` path. This not only means that we are including extra assets that we don't want depending on the section, but it also means we are overriding styles and functions for other projects and this may not be desired.

From our end we'll make sure to move our extensions angular modules into their own path when it makes sense and also add prefixes to avoid clashes.

PR #360, opened 2020-02-13
REF Code cleanup

mattwire:
- Regenerate boilerplate/entities and drop support for CiviCRM < 5.19
- Convert to short array syntax
- Update composer libraries for PHP7.3 support


mattwire:
Note #363 now covers the DAO update

PR #361, opened 2020-02-13
Allow extensions to enable smarty in the tokenprocessor context

mattwire:
For example:
```
  Civi::dispatcher()->addListener('civi.flexmailer.compose', 'myextension_symfony_civicrm_flexmailer_compose', \Civi\FlexMailer\FlexMailer::WEIGHT_PREPARE);
```
```
function myextension_symfony_civicrm_flexmailer_compose($event, $hook) {
  $event->context['smarty'] = TRUE;
}
```

### DIFF
--- a/CRM/Mosaico/MosaicoComposer.php
+++ b/CRM/Mosaico/MosaicoComposer.php
@@ -18,8 +18,8 @@ class CRM_Mosaico_MosaicoComposer extends \Civi\FlexMailer\Listener\DefaultCompo
     \Civi\FlexMailer\Event\ComposeBatchEvent $e
   ) {
     $context = parent::createTokenProcessorContext($e);
-    // Smarty would break Mosaico CSS.
-    $context['smarty'] = FALSE;
+    // Smarty would break Mosaico CSS unless we know what we are doing (eg. handling via extension)
+    $context['smarty'] = $e->context['smarty'] ?? FALSE;
     return $context;
   }
 


CRM/Mosaico/MosaicoComposer.php
```diff
@@ -18,8 +18,8 @@ public function createTokenProcessorContext(
     \Civi\FlexMailer\Event\ComposeBatchEvent $e
   ) {
     $context = parent::createTokenProcessorContext($e);
-    // Smarty would break Mosaico CSS.
-    $context['smarty'] = FALSE;
+    // Smarty would break Mosaico CSS unless we know what we are doing (eg. handling via extension)
+    $context['smarty'] = $e->context['smarty'] ?? FALSE;
```
totten, line 22: If one really wants Smarty (per se) plus Mosaico, then I think this should be a different `template_type`.

1. Do you have a link for the (draft?) extension which handles this well? (Maybe that makes it clearer.)

2. Just to make sure I'm interpreting this correctly: The ultimate value of `$context['smarty']` does *not* stem from `CIVICRM_MAIL_SMARTY` (like in `parent::createTokenProcessorContext()`). This patch implies a new/different flag in `ComposerBatchEvent::$context['smarty']` which determines the activation, and the only way to toggle that is to tap into `civi.flexmailer.compose`. So... there's no way existing deployments would inadvertently flip this behavior on? (So, in that way, the change is safe.)

3. In the same way that `*.tpl` and `*.twig` and `*.jsp` are different file formats, this feels like a different format. Data written in one format (e.g. `*.twig` -- or Mosaico with CiviMail templating) is not interchangeable with data written in the other format (e.g. `*.tpl` -- or Mosaico with Smarty templating) even if they look similar (HTML tags).

4. The Smarty/CSS issue is the most painful problem with `CIVICRM_MAIL_SMARTY` or `$context['smarty'] = TRUE`, but IMHO the semantics of that flag are broken by design. Not because Smarty is bad language -- but because the hybrid of CiviMail-Evaluation-then-Smarty-Evaluation is a needlessly janky way to use it (*broken compiler cache; needing to understand the interaction of multiple languages*). Template evaluation should be atomic/singular (not daisy-chained). There's obviously a pressure/inertia to use it anyway (less ΔSLOC; some users familiar with `CIVICRM_MAIL_SMARTY`). Personally, I'm in a mindset of "avoid that flag at all costs!".

mattwire:
@totten Thanks for reviewing and sorry for coming back so late! The draft extension is here: https://github.com/mjwconsult/civicrm-mosaico_smarty

I think I agree that it would be a good idea to extend the messagetemplate object to have a `template_type` and see if we can manage without the smarty context flag.

mattwire:
@totten I agree with your comments and intend to work through the `template_type` / `renderer` issues for message templates with a view to adding twig as an option too. For now I'm going to merge this because 1) it does not change the status-quo unless you specifically know what you are doing, in which case you should be ok with any errors it produces. 2) it means you don't need to patch mosaico if you do want to use smarty. 3) It's going to be a few months at least before we have all the necessary code in place for template_type/renderer etc.

PR #374, opened 2020-04-28
Add API method to update template URLs

elcapo:
When a CiviCRM database dump that contains Mosaico templates is restored in a server with a different URL, this method facilitates updating the templates:

```bash
cv api MosaicoTemplate.replaceurls from_url="http://old.server.org" to_url="https://new.server.org"
```


mattwire:
@elcapo Can you document this here: https://github.com/veda-consulting/uk.co.vedaconsulting.mosaico/blob/2.x/docs/api.md

Note there is currently something here: https://github.com/veda-consulting/uk.co.vedaconsulting.mosaico/blob/2.x/docs/index.md#migration-to-a-new-domain

elcapo:
@mattwire Done. Let me know if it looks correct now.

PR #388, opened 2020-08-03
adds bullets (unorderd list) to editor

AlainBenbassat:
A bulleted list is very common in mailings.

This pull request adds this option in the toolbar of the editor.

### DIFF
--- a/CRM/Mosaico/Page/Editor.php
+++ b/CRM/Mosaico/Page/Editor.php
@@ -98,7 +98,7 @@ class CRM_Mosaico_Page_Editor extends CRM_Core_Page {
       ],
       'tinymceConfigFull' => [
         'plugins' => ['link hr paste lists textcolor code civicrmtoken'],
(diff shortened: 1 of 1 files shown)


mattwire:
@AlainBenbassat I think this has had push back before because technically it can be done via a hook. However I'm of the opinion that it should be included as standard per this PR so if there are no objections I'll merge in a week or so. @MikeyMJCO @eileenmcnaughton

AlainBenbassat:
@mattwire I agree it's a good trade-off to add it in code because ordered lists - like bold and hyperlinks - are so common in digital communication / mailings that creating an extension for this is somewhat overkill.
@eileenmcnaughton @MikeyMJCO

eileenmcnaughton:
Hmm - so I don't feel best qualified to comment on this - but if it's just a case of enabling a common, standardised useful option I don't see a problem

Issue #319, opened 2019-07-16
Easy to misclick away and lose saved progress on WordPress


simonjohnparker:
Hello,

On Wordpress, running into a frustrating issue quite a lot.

When modifying a template, it's so easy to accidentally click the top bar:
<img width="1680" alt="Screenshot 2019-07-16 at 13 49 38" src="https://user-images.githubusercontent.com/43149642/61295840-d6c85480-a7d0-11e9-929f-8a31d7460ae2.png">

This results in navigating away from Mosaico and losing any unsaved progress, as it doesn't appear autosave exists.

On the usual WP pages/posts even though it autosaves, you still get this warning when you accidentally click to navigate away:
<img width="1680" alt="Screenshot 2019-07-16 at 13 42 39" src="https://user-images.githubusercontent.com/43149642/61295971-142ce200-a7d1-11e9-9d31-326f488130b5.png">

Is it possible to implement something similar with Mosaico?

Although you can disable the top black bar, I don't' want to do that as it's useful the rest of the time.

Upperholme:
I've had the same issue with Drupal - putting a newsletter together, move the mouse to the 'Blocks' menu item, overshoot slightly which causes the 'Search menu to drop down, and instead of clicking on 'Blocks' I've clicked on 'Find Contacts' and a lot of work is immediately lost. A pop-up alert as suggested would save the day. I've certainly lost several hours work through this, combined with the lack of autosave.

johncronan:
It happened to me within 20 minutes of starting to use the extension!

mwestergaard:
Been there, done that!

Barijohn:
We've also had client request something like this to stop them navigating away from the page. 

Or could a Save button be added instead of having to Close the page down to save the work done.

colemanw:
Looks like this issue can be closed @mattwire

PR #391, opened 2020-08-13
Warn user before clicking away from mosaico editor

mattwire:
This pops up a "browser" warning - "Are you sure you want to navigate away from this page?" using the standard `window.onbeforeunload` function. The custom message may or may not display depending on browser (it did not in Chrome 84).

@bhumanists @petednz @agileware-justin @artfulrobot @deepak-srivastava please test and let me know if this is more helpful for your users in or out!

### DIFF
--- a/templates/CRM/Mosaico/Page/EditorIframe.tpl
+++ b/templates/CRM/Mosaico/Page/EditorIframe.tpl
@@ -27,7 +27,7 @@
 
       window.onbeforeunload = function(e) {
         e.preventDefault();
-        e.returnValue = "Do you really want to leave our brilliant application?";
+        e.returnValue = "{/literal}{ts}Exit email composer without saving?{/ts}{literal}";
       };
 
       if (window.top.crmMosaicoIframe) {
--- a/templates/CRM/Mosaico/Page/EditorIframe.tpl
+++ b/templates/CRM/Mosaico/Page/EditorIframe.tpl
@@ -24,6 +24,12 @@
 
       var plugins = [];
       var config = {/literal}{$mosaicoConfig}{literal};
+
+      window.onbeforeunload = function(e) {
+        e.preventDefault();
+        e.returnValue = "Do you really want to leave our brilliant application?";
+      };
+
       if (window.top.crmMosaicoIframe) {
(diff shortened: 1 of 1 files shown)


bhahumanists:
Yes, I get the message correctly. With Chrome 84 I don't get a custom message, but the default is fine. 

Thanks! This should be very helpful.

mattwire:
Fixes #319

templates/CRM/Mosaico/Page/EditorIframe.tpl
```diff
@@ -24,6 +24,12 @@
 
       var plugins = [];
       var config = {/literal}{$mosaicoConfig}{literal};
+
+      window.onbeforeunload = function(e) {
+        e.preventDefault();
+        e.returnValue = "Do you really want to leave our brilliant application?";
```
colemanw: This message is funny but could be a little more informative. And translated. How about this:

```suggestion
        e.returnValue = "{/literal}{ts}Exit email composer without saving?{/ts}{literal}";
```

colemanw: Actually, I'm not sure because I haven't tried it... does this message pop up regardless of unsaved changes? Does it even pop up when exiting the application normally? That might be a problem.

mattwire: Haha! The message doesn't actually appear at all in most modern browsers which is why I didn't notice I'd left it per an example I found... Totally agree on updating it per your suggestion. It won't popup when exiting normally because it's triggered by the browser when clicking something outside the context of the form/window that will trigger it to disappear.

larssandergreen:
What needs to happen to add the nuance here so that the warning doesn't show up every time someone tries to save the mailing draft and exit the mailing (as discussed in #400 )?
We're having to train users to ignore the warning, but then we're also running into some other issues (like the "Mailing has not been saved" error), which users are also ignoring, quite reasonably. It's becoming a bit of a problem and it gets a bit ridiculous to try to teach users to ignore this warning, but don't ignore that other similar warning and so on.

larssandergreen:
@mattwire What about just adding window.onbeforeunload = null for clicks on the Save Draft and Submit buttons?

PR #321, opened 2019-07-23
Introduce A/B testing for subject lines

totten:
Before
=====

Creating an A/B test with Mosaico content is pretty cumbersome -- e.g. make a standalone Mosaico template; then enable the msg-tpl sync; then make an A/B test; and in the content of the test, use the msg-tpl; and don't try to change any of the content in the tpl.

After
====

In the "New Mailing" screen, the subject line shows a "+" button.

Screen Shot 2019-11-25 at 8 38 31 PM

Clicking the "+" button will create a few more fields -- for the alternative subjects and the sample sizing.

Screen Shot 2019-11-25 at 8 39 10 PM

Comments
============

Some interesting bits:

* The composition screen uses Mosaico UI - and content can be rearranged in-situ with Mosaico
* This should work in `crm-*` and in Shoreditch.
* When you send a test mailing, you'll get two emails (one with each variant)
* When you submit/finalize the Mosaico blast, it is converted to a formal CiviMail A/B test. The standard CiviMail UI already provides a set of links/reports/workflows for tracking A/B test mailings, and these will work the same on the Mosaico A/B mailings.

PR #403, opened 2020-09-28
Restore gulp3. Regenerate CSS.

totten:
The `gulp` command builds the CSS.  It is currently broken.  The package was upgraded to v4 as part of a long list of updates (13160c633e12471df7e78a8e3c3998ba1c6a71d6), but the current `gulpfile.js` only works on v3.

Just to demonstrate workingness (and to potentially facilitate debugging if there are any quirks), I've run `gulp` again to regenerate the CSS.


mattwire:
Merging as this is CSS only and won't be "seen" on sites until a new release is out. Looks like this will get some testing via #321

PR #406, opened 2020-10-30
placeholder images not accessible in joomla

deepak-srivastava:
Since placeholder images were permissioned, placeholder images have gone missing for Joomla installs.

This is mainly because in Joomla backend backend url (example.org/administrator) is separate from frontend url (example.org/index.php) and civicrm/mosaico/img url is frontend. See following screenshot:

image

The issue was also discussed [here](https://github.com/veda-consulting-company/uk.co.vedaconsulting.mosaico/issues/340#issuecomment-550557572). 

UPDATE:
- PR creates a new separate url for placeholder images which is populated as a backend url with it's own permissions - civicrm/mosaico/img/placeholder
- Also dependent on PR to mosaico library - https://github.com/civicrm/mosaico/pull/6

### DIFF
--- a/CRM/Mosaico/Utils.php
+++ b/CRM/Mosaico/Utils.php
@@ -326,16 +326,6 @@ class CRM_Mosaico_Utils {
 
       switch ($method) {
         case 'placeholder':
-
-          // Only privileged users can request generation of placeholders
-          // Skip permission check for Joomla as backend base url (example.org/administrator) is separate from
-          // frontend url (example.org/index.php) and civicrm/mosaico/img url is frontend.
-          if (CRM_Core_Config::singleton()->userFramework != 'Joomla' &&
-            !CRM_Core_Permission::check([['access CiviMail', 'create mailings', 'edit message templates']])
-          ) {
-            CRM_Utils_System::permissionDenied();
-          }
-
           Civi::service('mosaico_graphics')->sendPlaceholder($width, $height);
           break;
 
--- a/CRM/Mosaico/Utils.php
+++ b/CRM/Mosaico/Utils.php
@@ -328,7 +328,11 @@ class CRM_Mosaico_Utils {
         case 'placeholder':
 
           // Only privileged users can request generation of placeholders
-          if (!CRM_Core_Permission::check([['access CiviMail', 'create mailings', 'edit message templates']])) {
+          // Skip permission check for Joomla as backend base url (example.org/administrator) is separate from
+          // frontend url (example.org/index.php) and civicrm/mosaico/img url is frontend.
+          if (CRM_Core_Config::singleton()->userFramework != 'Joomla' &&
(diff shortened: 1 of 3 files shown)


mattwire:
I'm not able to suggest an alternative but I do have concerns with this as I do see regular attempts to abuse this endpoint on a couple of (wordpress) client sites. I guess the issue is that the placeholder URL needs to be frontend when viewing the email but not when composing? Could we do some postprocessing on the template instead to work around this?

aydun:
Back [here](https://github.com/veda-consulting-company/uk.co.vedaconsulting.mosaico/issues/340#issuecomment-549601448) @deepak-srivastava said:

> Now civicrm/mosaico/img is a <is_public> url and therefore generates a frontend url like:
> https://example.org/index.php?option=com_civicrm&task=civicrm/mosaico/img&method=placeholder&params=166%2C130

Since the placeholder images aren't intended to be in public mails and cause some DoS concerns, how about creating a civicrm/mosaico/admin-img for placeholder images with permissions and leaving truly public images in the current location?

deepak-srivastava:
@aydun thanks. Initially i wasn't sure how difficult it would be because of interference with mosaico library. But turns out it's not that difficult, as library already has a separate section for placeholder images - https://github.com/civicrm/mosaico/pull/6.

@mattwire 
PR has been updated to use a new placeholder image url with it's own permission and exposed as backend url.
Also see dependent PR https://github.com/civicrm/mosaico/pull/6.

eileenmcnaughton:
Seems OK to me ... @aydun @mattwire ?

CRM/Mosaico/Utils.php
```diff
@@ -326,12 +326,6 @@ public static function processImg() {
 
       switch ($method) {
         case 'placeholder':
-
-          // Only privileged users can request generation of placeholders
-          if (!CRM_Core_Permission::check([['access CiviMail', 'create mailings', 'edit message templates']])) {
-            CRM_Utils_System::permissionDenied();
-          }
-
```
mattwire, line 334: Did you check that the xml access_arguments mean we don't need the check here?

mattwire:
This seems ok yes, just one comment for @deepak-srivastava to confirm

CRM/Mosaico/Utils.php
deepak-srivastava: Shouldn't require the check in php for sure.
I did test it for Joomla which seemed to work due to backend url login prompt. And I just tried it for drupal to confirm, and it wasn't working as expected. 
Unlike other urls in mosaico menu xml file, it seems to require a access_callback, may be because in the tree mosaico/img url is public, and is somehow being derived (as public).
Have explicitly set the access_callback and tested. Seems to work now. Don't need the check in php.
Thanks for the headsup.

deepak-srivastava:
@mattwire do you need anything else?

mattwire:
@deepak-srivastava this seems good. Could just do with someone confirming they've run and tested this.

PR #409, opened 2020-11-20
Fix Leave site? warning when not required

larssandergreen:
Before: Leave site? warning was triggered by onbeforeunload when saving the mailing draft or submitting the mailing and otherwise leaving the mailing (after the user had edited the design of the mailing).

After: Leave site? warning is only triggered when leaving the mailing while editing the mailing design in Mosaico and not in the CiviCRM parts of the mailing editor.

This change just checks if the Mosaico iframe is shown on the mailing editing page before triggering the Leave site? warning and does not trigger the warning if the iframe is not currently shown.

### DIFF
--- a/templates/CRM/Mosaico/Page/EditorIframe.tpl
+++ b/templates/CRM/Mosaico/Page/EditorIframe.tpl
@@ -26,7 +26,7 @@
       var config = {/literal}{$mosaicoConfig}{literal};
       
       window.addEventListener('beforeunload', function(e) {
-        if(window.parent.document.getElementsByTagName('iframe')[0].style.display !== "none") {
+        if(window.parent.document.getElementById('crm-mosaico').style.display !== "none") {
           e.preventDefault();
           e.returnValue = "{/literal}{ts}Exit email composer without saving?{/ts}{literal}";
         }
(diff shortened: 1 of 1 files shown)


templates/CRM/Mosaico/Page/EditorIframe.tpl
```diff
@@ -24,11 +24,13 @@
...
       var plugins = [];
       var config = {/literal}{$mosaicoConfig}{literal};
-
-      window.onbeforeunload = function(e) {
-        e.preventDefault();
-        e.returnValue = "{/literal}{ts}Exit email composer without saving?{/ts}{literal}";
-      };
+      
+      window.addEventListener('beforeunload', function(e) {
+        if(window.parent.document.getElementsByTagName('iframe')[0].style.display !== "none") {
```
colemanw: This selector would be simpler after #411 is merged

colemanw: It's merged. Now we can change this to
```suggestion
        if(window.parent.document.getElementById('crm-mosaico').style.display !== "none") {
```

larssandergreen:
Much better, works for me

PR #411, opened 2020-12-01
Add dom id to mosaico iframe

colemanw:
This makes the element easier to work with, for example crmTutorial needs an id to access the contents with.


larssandergreen (commented):
Ignore, comment in wrong place

PR #412, opened 2020-12-04
Use standard methods and add validation to URL params for placeholder image endpoint

mattwire:
This improves validation on URL params for the placeholder image endpoint on Mosaico.

A number of sites are seeing attempts to abuse this endpoint and this reduces the attack vector.


mattwire:
@deepak-srivastava @seamuslee001 Could you please take a look?

deepak-srivastava:
Looks good to me :+1:

seamuslee001:
I think this makes sense but want r-run it first

mattwire:
Merging so this can be tested via the automated builds

Issue #241, opened 2018-02-28
Feature Request: Hide Base Templates


themak1985:
We have set up one template for our users - that one template meets our needs.  However - there are the three base templates + the template we created on display for users.  Is it possible to either choose what templates are displayed (ideally) or alternatively to hide base templates from non-admins.  

As you can see from this [Stack Exchange](https://civicrm.stackexchange.com/questions/22832/hide-base-templates-in-mosaico/22858#22858), other users have found their own ways with dealing with the templates.

totten:
To improvise a downstream fix:
 * I agree that deleting an unused base-template is pretty quick/simple.
 * One could probably implement `hook_civicrm_mosaicoBaseTemplates` and remove any unused base-templates.
 * One could probably inject some CSS. (Or, with a small patch to the Angular HTML, make the CSS possible.)

However, those are improvised downstream fixes. I can see how this might be a more common problem. 

Just to verbalize some designs, maybe....

* Allow some kind of extra permissioning -- such that users with more advanced permissions can see more templates? (Trick: *How to make this work without breaking existing installs?*)
* In the management screen for Mosaico templates, change the "Settings' to a bigger dialog. Provide some options for activating/deactivating each one.

lcdservices:
FYI, I was looking to do this as well and tried option 2 -- unsetting some of the templates using `hook_civicrm_mosaicoBaseTemplates` -- and that throws console errors: "TypeError: Cannot read property 'title' of undefined". I suspect the list of templates is loaded in another way as well, so even though you remove it here, the js is cycling through the expected list and throwing errors when it can't find it.

francescbassas:
I am a supporter of directly removing the default base templates from the extension. In any case put them in another extension.

In case of keeping them, leave them hidden by default or at most make visible the versafix that I guess is the only one that is really being used.

howardshand:
Was just setting this extension up and was looking for an option similar to the second design @totten suggests. Any chance of a revisit of this feature request.

For others the workaround we implemented was:
[...]

PR #414, opened 2020-12-07
Add setting to hide base templates

elcapo:
This is a merge request that attempts to fix the problem described in issue #241. In particular, it adds a new setting called `mosaico_hide_base_templates` that allows setting a list of template ids to be hidden. In the screenshot, the setting is being used to hide 2 of the 3 base templates:

Screenshot_2020-12-07 Mosaico Settings

**Important Note:**

When hiding a template, also messages that were created using it won't be shown. Therefore, some use cases are:

1. if you're only using `versafix-1`, you can safely disable `tedc15` and `tutorial`,
2. if you're only using a custom template (as [CiviVersafix](https://lab.civicrm.org/MikeyMJCO/CiviVersafix/)), you can also safely disable `versafix-1`.

If you decided to hide the 3 templates and you don't install a custom one, you won't leave any available template to work with.


mattwire:
@elcapo Does this also work with the custom templates directory?

elcapo:
> @elcapo Does this also work with the custom templates directory?

Sorry for the late answer. Yes, it does!

elcapo:
I've also switched the UI to a slightly more usable multiple selector:

Screenshot from 2021-01-28 13-18-19

Also, to facilitate the revision of the changes, it might be useful to check the `findBaseTemplates` method of the `CRM_Mosaico_BAO_MosaicoTemplate` class, as I added two parameters to it.

By default, it works as it always did. But by using those parameters we can now:

- call the method skipping the cache by passing it the first `$ignoreCache` as true
- disable the hooks execution by passing it the second `$dispatchHooks` as false

This is important because what the extension is doing here is basically to alter the way it creates the list of base templates:

- when using it to show the templates in the mailing editor, it will work as always,
- but when using the same method (with the parameters described above) for the options of templates to hide, it will also show the hidden ones.

homotechsual:
This is working great for me in testing. I like the multiselect UI option.

PR #426, opened 2021-01-09
Update AutoGenerated DAO, Civix, auto_install files to be latest vers…

### LINKED ISSUE #425: Array and string offset access syntax with curly braces is deprecated
From PHP 7.4 uk.co.vedaconsulting.mosaico throws al lot of these notices. 
Not harmful but probably good to take in account for a new version.

seamuslee001:
…ions from civix commands

ping @mattwire should also resolve #425


sql/auto_install.sql
```diff
@@ -1,23 +1,90 @@
...
+     `hash_key` varchar(32) NOT NULL   ,
+     `name` varchar(32) NOT NULL   COMMENT 'name',
+     `html` longtext NOT NULL   COMMENT 'HTML',
+     `metadata` longtext NOT NULL   COMMENT 'metadata',
+     `template` longtext NOT NULL   COMMENT 'template' 
+,
+        PRIMARY KEY (`id`)
+ 
+ 
+,          CONSTRAINT FK_civicrm_mosaico_msg_template_msg_tpl_id FOREIGN KEY (`msg_tpl_id`) REFERENCES `civicrm_msg_template`(`id`) ON DELETE CASCADE
```
mattwire, line 63: I generally add `ENGINE=InnoDB` because otherwise I get quite a few complaints as apparently some shared servers are not setting that as default - including "civihosting" which seems particularly irresponsible but anyway..

seamuslee001: Added in now @mattwire

eileenmcnaughton:
Looks like the issue was addressed & this is basic boilerplate (& will remove a deprecation notice) - merging

PR #433, opened 2021-02-23
Throw error message on resizing image and prevent hard fail

monishdeb:
Mosaico uses the Intervention library to handle image manipulation. we've found a few instances where this throws a fatal error and breaks the application, when we think it should log an error to file, perhaps throw a  UI warning -- but not block use of the system. Two cases where it causes hard failure:
1.  Files in the images/uploads folder (i.e. the Mosaico gallery) had improper permissions, and
2. when there were files there that were not valid image files.

BEFORE
---------
Hard failure breaks the design screen

AFTER
-------
Capture the error messages and notify the user on compose screen.


lcdservices:
This works as expected. The UI can still be used.

eileenmcnaughton:
Seems OK to me - @mattwire ?

mattwire:
This seems sensible and has been tested by @lcdservices

PR #442, opened 2021-04-09
show the base templates that are not hidden

### LINKED ISSUE #241: Feature Request: Hide Base Templates

kurund:
* Fixes: #241


CRM/Mosaico/BAO/MosaicoTemplate.php
```diff
@@ -72,6 +72,9 @@ public static function findBaseTemplates($ignoreCache = FALSE, $dispatchHooks =
         }
       }
 
+      // get list of base templates that needs be to hidden from the UI
+      $templatesToHide = CRM_Core_BAO_Setting::getItem('Mosaico Preferences', 'mosaico_hide_base_templates');
```
seamuslee001: @kurund can we use \Civi::settings()->get('mosaico_hide_base_templates') here ?

kurund: Thanks for reviewing. I have made the changes.

seamuslee001:
I tested this and confirmed that it fixes the issue reported by Krund and works for me @mattwire

PR #445, opened 2021-04-19
Specify to use bootstrap3 bundle if in bootstrap mode

seamuslee001:
ping @mattwire @colemanw @totten This is to be consistent with https://github.com/civicrm/civicrm-core/blob/5db0bc3c1f54eaca4307f103a73bda596ae914d6/ext/search/ang/crmSearchDisplayList.ang.php#L13


mattwire:
@seamuslee001 This implies a minimum CiviCRM core version (I can't remember which one bundles were added). My "policy" is to set the minimum version to the last security release - can you update info.xml to min 5.35 to ensure this doesn't get installed on versions that don't support bundles.

seamuslee001:
Thnaks @mattwire good point have updated that now

Issue #427, opened 2021-01-11
Class \CRM_Mosaico_AbDemux not found - potential conflict with org.civicoop.areas ?


ufundo:
Hi,

Upgrade from 2.5 to 2.6 caused this fatal error around the container / service loading for me on CiviCRM 5.28.3 , Drupal 7, PHP 7.3:

`Class \CRM_Mosaico_AbDemux not found in CachedCiviContainer->getMosaicoAbDemuxService() (line 840 .../civicrm/templates_c/CachedCiviContainer.~~~~.php)`

Tried clearing caches and manually emptying templates_c directory with no success. Works again downgrading to 2.5. Some compatibility issue?

eileenmcnaughton:
@totten @seamuslee001 are we also clearing the services on extension upgrades in an ok order?

@ufundo if you manually delete the file above it should work? The services are cached in files like ``CachedCiviContainer``

ufundo:
Hi @eileenmcnaughton thanks for help.

Was getting the same error even after deleting the `CachedCiviContainer` files ( and everything else in templates_c ). I think maybe it had crashed/corrupted the extension upgrade process.

I noticed this https://github.com/civicrm/civicrm-core/pull/19141 and wondered if might be relevant. Have just found the extension upgrade seems to work OK if the site is upgraded to 5.33.1 before extension upgrade.

If the container-thing-I-don't-understand is required for smooth upgrade perhaps this new 2.6 Mosaico release should require 5.32 where it is merged into core?

eileenmcnaughton:
@mattwire I think that makes sense to increment the supported version to reduce issues people hit / have to be supported - people on 5.28 can continue to use 2.5 until the next civi security release drops...

mattwire:
@ufundo @eileenmcnaughton I've upgraded a number of 5.28.4 sites to 2.6 with no issues. That said, I had similar issues on a number of sites when flexmailer moved from extension to core extension and #19141 probably resolves those issues too. So I'd say not mosaico specific. I don't mind incrementing min-version but 2.6 is already released and published.

magnolia61:
@ufundo I have the same issue as you do. Cant upgrade to 2.6 without the fatal error you describe

ufundo:
Sorry @magnolia61 (though I guess it's nice to know it's not *just* me ;)

[...]

PR #446, opened 2021-04-23
Fix for issue #427

jaapjansma:


mattwire:
@totten This look ok?

totten:
Aaah, very interesting. I've only skimmed the original discussion, but I'm OK with this conceptually as a near-term fix for Mosaico.

I would be curious about whether it's been fixed incidentally on `civicrm-core@master`. We had some recent PRs to address other "class not found" issues for edge-cases of the extension-lifecycle.

* https://github.com/civicrm/civicrm-core/pull/19870
* https://github.com/civicrm/civicrm-core/pull/20091
* https://github.com/civicrm/civicrm-core/pull/20116

Specifically, if one adds this to `info.xml`, then it should provide more robust classloading:

```xml
<classloader>
  <psr0 prefix="CRM_" path="" />
</classloader>
```

jaapjansma:
I have tried to add the code below to `info.xml` but the error still occurs. (It is on CiviCRM 5.35.0)

```
<classloader>
  <psr0 prefix="CRM_" path="" />
</classloader>
```

PR #463, opened 2021-06-29
Improve error handling on image params

mattwire:
Cleaner handling of method/params parameters for image processing endpoint. If parameters are invalid log an error and return bad request.

Before it was inconsistent and there seems little point throwing an exception or fatal error because the endpoint is generally used over AJAX.

@seamuslee001

### DIFF
--- a/CRM/Mosaico/Utils.php
+++ b/CRM/Mosaico/Utils.php
@@ -321,12 +321,19 @@ class CRM_Mosaico_Utils {
       CRM_Utils_System::civiExit();
     }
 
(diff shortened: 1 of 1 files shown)


seamuslee001:
So if this regex doesn't get matched https://github.com/veda-consulting-company/uk.co.vedaconsulting.mosaico/blob/2.x/CRM/Mosaico/ImageFilter.php#L37 it would still be used on the final image usage which wouldn't be over AJAX but I think that this is probably sensible none the less, however I would also think that you may want to adjust the other 2 Exceptions in this function as well?

PR #468, opened 2021-10-06
Allow users to choose hotlist tokens

jitendrapurohit:
### Before 

Hotlist tokens are fixed.

image

### After

Can be configured in settings

image

image


totten:
+1 on concept.

I think I had started testing this a couple weeks ago and got side-tracked, so nothing definitive to add. Something to clarify is the sorting - eg whether tokens are presented in (a) the order they are chosen, (b) alphabetical by label (`First Name`), (c) alphabetical by machine name (`{contact.first_name`), or (d) some other ordering.

mattwire:
This seems to work fine. The sort-order seems to be passed to javascript in label alphabetical but then looks like it's displayed in alphabetical token/machine name.

I don't think that's a big problem as we're only showing a short list. But it could be fixed in a follow-up PR if required.

Issue #444, opened 2021-04-17
Autosave - possible Make it Happen?


larssandergreen:
I think quite a few folks have run into frustrations with lost work in Mosaico mailings. Some work has been done to prevent users from clicking away by accident, but I think an autosave feature that saves the mailing as you work would probably be very helpful for many (and is becoming pretty standard these days, so users are frustrated to find that Mosaico doesn't do this).

Would someone be willing to take this on? Does anyone have a rough sense of how much work this would be? If it is reasonable, I think this could be a fundable Make it Happen. We would certainly kick in for this and there is at least one other [org who are interested](https://civicrm.stackexchange.com/questions/38675/auto-save-mailing-editing-template-in-mosaico). This is probably the biggest problem we have with CiviCRM in our org right now and I'm sure we're not alone.

Depending on how big of a project this is, it could possibly also be combined with work in core to lock mailings when they are being edited by another user or at least to warn that another user has a mailing open for editing, [as I've suggested here.](https://lab.civicrm.org/dev/mail/-/issues/85). Together, I think these two improvements would save a couple days of staff time per year in our org alone.

mattwire:
This is probably of interest to @bhahumanists as well. Autosave already works on the template editor I think? In order to implement and test this properly I would estimate between 4-8 hours work. If setting up a Make it Happen it would probably be a good idea to extend the scope (ie. +10-16 hours) to upgrade the mosaico library to latest version as well.

larssandergreen:
That seems reasonable. We might even just be able to skip the Make it Happen and just do the autosave itself split between a few orgs. Can you share a cost estimate in dollars? Email is in my profile here if you prefer.

mattwire:
@larssandergreen Thankyou I have sent you an email

petednz:
i would chip in to help

larssandergreen:
@petednz Great! Matt is going to scope this soon and then I'll get the MiH up.

petednz:
@mattwire any idea on timeframe for this?

mattwire:
> 1) Auto-save in Mosaico editor
> Prominent, persistent warning if auto-save stops working, recommending closing and re-opening the editor

MJW:
Currently autosave is not implemented. CiviCRM core has an implementation of "autosave" in the angular `CrmAutosaveCtrl` control - that autosaves the mailing details (every 5 minutes) but not the mosaico editor changes - they are only saved when you click close or test.
[...]

PR #471, opened 2021-10-21
(WIP; #444) Auto-save via periodic sync

totten:
Overview
--------------

Implement an auto-save mechanism while editing Mosaico mailings and templates.

This PR is marked "WIP" because it subtly changes the data-model and likely requires other updates to the UI (more details below).

Background
----------------

Recall that the "New Mailing" UI is a hybrid of two JS apps, the CiviMail screen (AngularJS) and the Mosaico screen (KnockoutJS). Generally, the CiviMail side is responsible for loading/saving data - but it launches Mosaico in an IFRAME. Eventually, the Mosaico IFRAME sends data back to CiviMail.

Mosaico presents a few APIs for exchanging data. Most notably for this PR:

* `viewModel.exportMetadata()` - Get very basic info (eg the name of the base-template). This method is essentially free (<1ms on i3-10100).
* `viewModel.exportJSON()` - Get the list of blocks and styling options that the user chose. This method is essentially free (<1ms on i3-10100).
* `viewModel.exportHTML()` - Get the rendered HTML. This is composition based on the JSON data and the base-template. It is fairly expensive (200ms-300ms on i3-10100).

AFAIK, these are the only public/supported APIs for accessing the content of the editor. If you want some other API (eg incremental updates for specific blocks or text content), then that requires internal development of Mosaico/Knockout. To have maintainable interaction, we have to use these methods.

Approach
------------

This PR updates `CrmMosaicoIframe` to support a periodic sync function, e.g.

```js
new CrmMosaicoIframe({
  syncInterval: 1000
  actions: {
    sync: function(ko, viewModel, iframeState) {
      ...viewModel.exportJSON()...
    }
  }
})
```

But what, exactly, should you do in `sync`? Call `exportJSON()`? Or `exportMetadata()`? Or `exportHTML()`? Ideally, you call all three and save the corresponding values to a `Mailing` or `MosaicoTemplate` record:

| Export from... | Save to (`Mailing`) | Save to (`MosaicoTemplate`) |
| -- | -- | -- |
| `exportMetadata()` | `Mailing.template_options.mosaicoMetadata` | `MosaicoTemplate.metadata` |
| `exportJSON()` | `Mailing.template_options.mosaicoContent` | `MosaicoTemplate.content` |
| `exportHTML()` | `Mailing.body_html` | `MosaicoTemplate.html` |

The problem is that `exportHTML()` is fairly expensive (eg 300ms). If the user is actively using the editor (eg editing text or dragging content around), then it's liable to cause hiccups. Anecdotally, it caused extra JS warnings and slight stuttering (while dragging). And you don't really *need* the HTML at this point - you're still editing, not sending.

So this PR does a partial save, ie the metadata and JSON content are saved. The HTML is left out-of-date, but it's flagged as stale (`<!--STALE-->`). If the browser crashes, you still come back and resume editing (based on the saved JSON).

Issues/Considerations
-------------------------

If the content of  `Mailing.body_html` or `MosaicoTemplate.html` includes the `<!--STALE-->` flag, then it is a valid reflection of *some iteration* of the content, but it is not an *update to date* iteration. This introduces new edge-cases that need to be identified/resolved.

__For `Mailing` records__: The mailing `body_html` isn't really used until you "Preview", "Send test", or "Submit", and all those actions require coming back to the CiviMail editor. So I think it'll be sufficient for the CiviMail editor to do some validation+warnings. The process would look like this:

1. Create a new mailing. Start editing in "Mosaico" iframe. Let it autosave. But force-close, such that we only have autosave content.
2. Navigate back to resume editing. The CiviMail UI detects the `<!--STALE-->` flag in `body_html`, so it shows a warning, and it prohibits you from using  "Preview" / "Send test" / "Submit". You'll need to reopen/resave in the Mosaico IFRAME.
    * __TODO__: Update the CiviMail (AngularJS) UI with these checks.

__For `MosaicoTemplate` records__: The `html` field is funny. I don't think it's actually consumed in this extension, but I believe it's used for https://lab.civicrm.org/extensions/mosaicomsgtpl. Consumers of `civicrm_mosaico_template` and `civicrm_msg_template` are a bit more open-ended (e.g. user Alice may be actively editing a `MosaicoTemplate` while user Bob sends a new message derived from the same `MosaicoTemplate`). Additionally, the user may not remember to come back and resave properly. So what do we do? These options stand out to me:

* Drop the `html` field. Break `mosaicomsgtpl`.
* Autosave the `html` with a `<!--STALE-->` flag (like in current draft). If some parallel process consumes `html`, it will get an `html` message with the invisible `<!--STALE-->` flag (ie old but cogent content)
    * __TODO__: If the user leaves prematurely, we're left with a `<!--STALE-->` record indefinitely We need a way to direct some user back to resume editing and to save properly. Probably a system-status-check?
    * __TODO__: `mosaicomsgtpl` might want to be a little more picky - eg don't pull `<!--STALE-->` content from `MosaicoTemplate.html`; wait until there's final content.

There's another limitation regarding:

__Active TinyMCE__: I haven't fully checked, but I get the impression that `viewModel.exportJSON()` does not give you the updated content of a TinyMCE block if it's being actively edited. I believe changing that would require upstream patches to `mosaico`. So for  `r-run`, one should explicitly look out for this. (Auto-saving *writ-large* is different from auto-saving *the active TinyMCE box*.) However, even with this limitation, it should still be much better than the status-quo.

### DIFF
--- a/ang/crmMosaico/Iframe.js
+++ b/ang/crmMosaico/Iframe.js
@@ -4,7 +4,7 @@
    * The class CrmMosaicoIframe allows you to instantiate and manage a
    * full-screen IFRAME with embedded Mosaico runtime.
    */
-  angular.module('crmMosaico').factory('CrmMosaicoIframe', function(crmUiAlert, $q, $timeout, $rootScope) {
+  angular.module('crmMosaico').factory('CrmMosaicoIframe', function(crmUiAlert, $q, $timeout, $rootScope, $interval) {
 
     /**
      * @param Object newOptions
@@ -22,6 +22,7 @@
     return function CrmMosaicoIframe(options){
       var cfg = {
         url: CRM.url('civicrm/mosaico/iframe', 'snippet=1'),
+        syncInterval: 1000,
         dimensions: function resize() {
           var c = CRM.crmMosaico || {};
           var top = 0, left = 0, width = $(window).width(), height = $(window).height();
@@ -40,6 +41,7 @@
 
       var model = cfg.model, actions = cfg.actions;
       var isVisible = false, $iframe = null, iframe = null;
+      var syncHandle = null;
 
       if (actions.save && actions.close) {
         throw "Error: Save and Close actions are mutually exclusive";
@@ -133,6 +135,10 @@
       };
 
       this.destroy = function destroy() {
+        if (syncHandle) {
+          $interval.cancel(syncHandle);
+          syncHandle = null;
+        }
         this.hide();
         $(window).off('resize', onResize);
         if ($iframe) $iframe.remove();
@@ -169,6 +175,11 @@
         }
         if (actions.test) {
           viewModel.test = mkCmd("Test", actions.test);
+        }
+        if (actions.sync) {
+          syncHandle = $interval(function() {
+            actions.sync(ko, viewModel, {isVisible: isVisible});
+          }, cfg.syncInterval);
         }
       }
 
--- a/ang/crmMosaico/ListCtrl.js
+++ b/ang/crmMosaico/ListCtrl.js
@@ -1,5 +1,11 @@
 (function(angular, $, _) {
 
+  // The STALE_FLAG is added to `body_html` when it is out-of-sync with the `template_options.mosaicoContent`.
+  // Why does this exist? We want to autosave while the user interacts with the editor.
+  // However, rendering HTML while the user interacts is problematic (eg requires ~200ms on i3-10100/Firefox v93;
+  // eg provokes JS warnings). So we save just `mosaicoContent` (JSON) and allow the `body_html` to grow stale.
+  var STALE_FLAG = '<!--STALE-->';
+
   angular.module('crmMosaico').config(function($routeProvider) {
       $routeProvider.when('/mosaico-template', {
         controller: 'CrmMosaicoListCtrl',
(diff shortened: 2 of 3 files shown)


totten:
I've confirmed the issue when autosaving with an "Active TinyMCE" editor. It could cause some confusion, and it reflects a design-decision in Mosaico - when editing a text box, changes will propagate when you click-away (aka "blur", aka "de-focus", aka "click somewhere else"). That may sound abstract - here's a [video to demonstrate it with rich text (tinymce) box](https://user-images.githubusercontent.com/1336047/138209819-a2b38db4-c4bc-44c5-8cbc-f6d7b572a962.mp4).

Hypothetically, if an author spent 10 minutes editing a big blob of prose, then Mosaico wouldn't acknowledge that edit until they click away to work on another paragraph - so it won't autosave during those 10 minutes. However, every other kind of edit (adding/moving/deleting blocks, inserting images, editing properties, etc) would autosave more promptly. Additionally, the templates encourage many small text boxes - if you compose in that style, then autosave will never be far behind. It's only noticeable if you spent a long time on a single block of text.

Which, of course, isn't great... but I don't think there's an easy solution (within the current API), and it's a lot better than the status quo.

(Edit/Recap) In short, if you exit the page incorrectly, you're vulnerable to losing *a-paragraph-worth-of-changes* rather losing *all changes*.

totten:
Unless anyone pipes up, let's close in favor of #472.

PR #472, opened 2021-10-22
(#444) Auto-save after idle-timeout (~4s) and/or active-timeout (~90s)

totten:
Overview
--------------

Implement an auto-save mechanism while editing Mosaico mailings and templates. This is alternative to #471.

Background
----------------

Recall that the "New Mailing" UI is a hybrid of two JS apps, the CiviMail screen (AngularJS) and the Mosaico screen (KnockoutJS). Generally, the CiviMail side is responsible for loading/saving data - but it launches Mosaico in an IFRAME. Eventually, the Mosaico IFRAME sends data back to CiviMail.

Mosaico presents a few APIs for exchanging data. Most notably for this PR:

* `viewModel.exportMetadata()` - Get very basic info (eg the name of the base-template). This method is essentially free (<1ms on i3-10100).
* `viewModel.exportJSON()` - Get the list of blocks and styling options that the user chose. This method is essentially free (<1ms on i3-10100).
* `viewModel.exportHTML()` - Get the rendered HTML. This is composition based on the JSON data and the base-template. It is fairly expensive (200ms-300ms on i3-10100).

AFAIK, these are the only public/supported APIs for accessing the content of the editor. If you want some other API (eg incremental updates for specific blocks or text content), then that requires internal development of Mosaico/Knockout. To have maintainable interaction, we have to use these methods.

Approach
------------

Like #471, this PR updates `CrmMosaicoIframe` to support a periodic sync function, e.g.

```js
new CrmMosaicoIframe({
  actions: {
    sync: function(ko, viewModel, iframeState) {
      ...viewModel.exportJSON()...
    }
  }
})
```

Each UI ("Mailings => New Mailing" and "Mailings => Mosaico Templates") uses this sync action to call `exportJSON()`/`exportHTML()` and propagate changes.

The scheduling of the `sync` method is somewhat special:

* The `sync` method never fires while there is an active key-down or active mouse-down (deferring until the key is released).
* The `sync` method fires after ~4s of idleness.
* Alternatively, the `sync` method fires after ~90s of near-continuous activity.

Limitations
-------------------------

__Active TinyMCE__: Mosaico does not update its internal model while a user is focused on a rich text box -- it updates when the user moves away from the textbox. Consequently, the autosave/sync/export processes do not have immediate visibility into pending changes in a paragraph. Autosave should still be useful in typical usage-patterns. [More discussion of this limitation](https://github.com/veda-consulting-company/uk.co.vedaconsulting.mosaico/pull/471#issuecomment-948251124).

__JS Console Warning__: When saving, it may sometimes generate a warning. I haven't found any particular breakage from it, but I can't decisively say whether it is safe or unsafe.

```
TODO exception scrolling into view TypeError: e is null
```

### DIFF
--- a/ang/crmMosaico/Iframe.js
+++ b/ang/crmMosaico/Iframe.js
@@ -4,7 +4,7 @@
    * The class CrmMosaicoIframe allows you to instantiate and manage a
    * full-screen IFRAME with embedded Mosaico runtime.
    */
-  angular.module('crmMosaico').factory('CrmMosaicoIframe', function(crmUiAlert, $q, $timeout, $rootScope) {
+  angular.module('crmMosaico').factory('CrmMosaicoIframe', function(crmUiAlert, $q, $timeout, $rootScope, CrmMosaicoSyncMonitor) {
 
     /**
      * @param Object newOptions
@@ -40,6 +40,7 @@
 
       var model = cfg.model, actions = cfg.actions;
       var isVisible = false, $iframe = null, iframe = null;
+      var syncMonitor = new CrmMosaicoSyncMonitor();
 
       if (actions.save && actions.close) {
         throw "Error: Save and Close actions are mutually exclusive";
@@ -116,6 +117,7 @@
       this.hide = function hide() {
         isVisible = false;
         if ($iframe) {
+          syncMonitor.stop($iframe);
           scrollRestore();
           $iframe.hide();
         }
@@ -128,6 +130,7 @@
           scrollHide();
           onResize();
           $iframe.show();
+          syncMonitor.start($iframe);
         }
         return this;
       };
@@ -170,6 +173,10 @@
         if (actions.test) {
           viewModel.test = mkCmd("Test", actions.test);
         }
+
+        syncMonitor.onSync = function () {
+          if (actions.sync) actions.sync(ko, viewModel);
+        };
       }
 
     };
--- a/ang/crmMosaico/ListCtrl.js
+++ b/ang/crmMosaico/ListCtrl.js
@@ -30,6 +30,16 @@
       }
     });
 
+    function _saveTpl(tplId, viewModel) {
+      viewModel.metadata.changed = Date.now();
+      return crmApi('MosaicoTemplate', 'create', {
+        id: tplId,
+        html: viewModel.exportHTML(),
+        metadata: viewModel.exportMetadata(),
+        content: viewModel.exportJSON()
+      });
+    }
+
     $scope.createTpl = function createTpl(tpl) {
       return crmMosaicoPrompt(ts('Create new template'), ts('New Template (%1)', {1: tpl.type}))
         .then(function(newTitle) {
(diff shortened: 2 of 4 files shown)


totten:
@monishdeb This PR was built after some feedback from testing/review on the earlier auto-save work. Any chance you could give it a review?

monishdeb:
Thanks, @totten for submitting the patch. I have applied the patch in my local Mosaico setup and works smoothly. Here's what I tried in the screencast mentioned below: 
1. Configured bulk mailing
2. Chosen template and started adding blocks. Noticed the different behavior on the action performed like mentioned in the description of how the sync method triggers autosave 
2.1 The template got saved 4secs of idleness 
2.2 Continous activity doesn't trigger autosave after a while, might be 90secs
2.3. Active keydown/mousedown (as per code [here](https://github.com/veda-consulting-company/uk.co.vedaconsulting.mosaico/pull/472/files#diff-66333a61a98478a10f0e4d7267831a1df6152eba3699f8efbd2481efe639a6f5R48)) skip trigerring autosave. 

3. Then without closing the template tab I went to the 'Draft and Unscheduled' mailing ... luckily the autosave was triggered just before switching the screen. 

4. Went back to the modified bulk mailing template, and confirmed that the changes were saved correctly. 

Screencast: 
before

The patch looks good to me. Btw I didn't encounter the JS Console Warning mentioned in the description. 

Not sure if it's a good idea, but is it possible to trigger the `sync` function to autosave the template (along with canceling the active session) whenever someone switches to other page, without clicking the 'Close' action on top right?

totten:
> The patch looks good to me. Btw I didn't encounter the JS Console Warning mentioned in the description.

Yay!

> Not sure if it's a good idea, but is it possible to trigger the sync function to autosave the template (along with canceling the active session) whenever someone switches to other page, without clicking the 'Close' action on top right?

That sounds really ideal to me, but we're not obligated (*the user is given a warning before they leave*), and I'm a little wary of the `beforeunload` event.

https://stackoverflow.com/questions/41909365/what-are-the-limitation-using-before-unload-event

> Any async code ([ED] *like XMLHttpRequest or setTimeout - both of which are in the current path of autosaving*)... will be queued on the event loop, but it may or may not finish executing depending on when your browser completes the "unload" action (e.g. close window, refresh page.

https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload

> To combat unwanted pop-ups, some browsers don't display prompts created in beforeunload event handlers unless the page has been interacted with. Moreover, some don't display them at all. 

So... it might work... sometimes?  I dunno - it feels like `beforeunload` is a point of contention in the arms-race between browser vendors and abusive sites... might be easier to keep it minimal.

lcdservices:
I did some testing and this worked well for me. The idea of saving before someone moves away from the screen would be great, but not a deal breaker. The patch is a huge improvement over what we currently have.

monishdeb:
I agree with @totten (thanks for the info about `beforeunload` limitation) and @lcdservices, not a blocker for this much needed improvement to get the template saved before switching screens. 


@deepak-srivastava @mattwire can you please review this PR ?

mattwire:
I have also tested and confirm it is working well

totten:
Yay

PR #476, opened 2021-12-08
Send test

### LINKED ISSUE #449: Sending test email to group: selects all contacts when there is no group selected
**Use-case**
After editing a template, I wanted to send a test email to my address to check the mail.
I accidentally clicked on "send test" to a group instead of "send test" to a single email. Then I was telled that "A test message will be sent to 14 people", which is all the contacts in the DB (this is a test instance).

That's fine, but on a production site there were 50,000 contacts one click away from receiving a test email and a very scared user...

Screenshot from 2021-05-07 18-19-22

**Setup**
Drupal 8.9.19
CiviCRM 5.43.2
Mosaico 2.8
Civi Theme: greenwich (I'm not convinced by shoreditch yet)

semseysandor:
fix #449 

**Before**
After debugging I found that in [`BlockPreview.js`](https://github.com/veda-consulting-company/uk.co.vedaconsulting.mosaico/blob/9ff7a27fdfb2ebecd77294a287a1ed057d9cedd8/ang/crmMosaico/BlockPreview.js#L33) if there is no group id supplied, it will list all contacts.

Then I realized this button shouldn't be enabled in the first place when there is no group selected. [`BlockPreview.html`](https://github.com/veda-consulting-company/uk.co.vedaconsulting.mosaico/blob/e95d72b03c2896004576f2b11e9be56e30370d0b/ang/crmMosaico.crmstar/BlockPreview.html#L52)

As I see these buttons are an `<a>` element in `ang/crmMosaico.crmstar/BlockPreview.html` for which the `"disabled"` attribute is not defined.

**After**
I changed them to `<button>` elements (as in `ang/crmMosaico.bootstrap/BlockPreview.html`) and it has the desired functionality, however full width was forced for these elements to have the same look as the `<a>` elements.

### DIFF
--- a/ang/crmMosaico.crmstar/BlockPreview.html
+++ b/ang/crmMosaico.crmstar/BlockPreview.html
@@ -11,12 +11,12 @@ Vars: mailing:obj, testContact:obj, testGroup:obj, crmMailing:FormController
     </div>
     <div ng-hide="!mailing.body_html">
       <div>
-        <button class="btn btn-primary" ng-disabled="crmMailing.$invalid" ng-click="doPreview('html')">{{:: ts('Preview as HTML') }}</button>
+        <button class="btn btn-primary full-width-force" ng-disabled="crmMailing.$invalid" ng-click="doPreview('html')">{{:: ts('Preview as HTML') }}</button>
       </div>
     </div>
     <div ng-hide="!mailing.body_html && !mailing.body_text" style="margin-top: 1em;">
       <div>
-        <button class="btn btn-primary" ng-disabled="crmMailing.$invalid" ng-click="doPreview('text')">{{:: ts('Preview as Plain Text') }}</button>
+        <button class="btn btn-primary full-width-force" ng-disabled="crmMailing.$invalid" ng-click="doPreview('text')">{{:: ts('Preview as Plain Text') }}</button>
       </div>
     </div>
   </div>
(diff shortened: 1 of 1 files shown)


colemanw:
Per html spec, a `<button>` element must have `type="button"` or it will be assumed to be `type="submit"`. We don't want that.

semseysandor:
@colemanw I've added the `type="button"` attribute to the buttons, if I understood correctly your suggestion.

colemanw:
Looks good to me

mattwire:
@semseysandor Please can you "rebase" the PR? So that we can merge

semseysandor:
@mattwire Yeah, sure.

colemanw:
So... fun fact I just learned from @totten - nearly every html template in this extension has an evil twin! So this file has a doppleganger named `ang/crmMosaico.bootstrap/BlockPreview.html` and that file has to be carefully updated with the same changes to keep them in sync.

I hate this so much. @mattwire as soon as you cut the next release I'm going to make a PR to delete all the duplicate templates and just always use bootstrap (which is now shipped in core so there's no reason to keep all these duplicate templates)

semseysandor:
They seem to be in "more-or-less" sync (indepedently of this PR) - HTML markup is a bit different, but otherwise yes.
With this PR they won't be in sync, but I think the number of differences will reduce :)

colemanw:
Ah, I see that they are already `<button>` elements, so yes this is actually an improvement toward sync.
@mattwire this is good to merge now.

PR #478, opened 2022-01-10
Add Mosaico Template Listing page based on Searchkit

monishdeb:


colemanw:
I've reviewed the code and it looks good :+1:

mattwire:
@monishdeb With the optionvalues (categories) as managed entities is it possible to remove the default ones?

mattwire:
I think there will also be a problem that adding required extensions won't be picked up on upgrades so we'd also need a system check or similar to ensure that we enable those extensions on upgrade.

colemanw:
> I think there will also be a problem that adding required extensions won't be picked up on upgrades so we'd also need a system check or similar to ensure that we enable those extensions on upgrade.

IMO that problem is not limited to this extension, and the system-check should be added to core. However I don't think it should be a blocker to this PR because only the new functionality added in this PR will be affected by missing dependencies; existing functionality is unaffected.

@mattwire here is a core PR for that check: https://github.com/civicrm/civicrm-core/pull/22464

colemanw:
> I think there will also be a problem that adding required extensions won't be picked up on upgrades so we'd also need a system check or similar to ensure that we enable those extensions on upgrade.

Kind of. Managed entities can be modified but I'd have to double-check whether they can be deleted. IMO a good compromise here would be to only create one default option (Newsletters), delete the second option from this PR and let site admins configure it themselves.
@monishdeb

monishdeb:
@colemanw @mattwire I have made the following changes: 
1. Removed the 2nd option and kept only 'Newsletter' as the default option in the mgd file.
2. Changed new page label to 'Mosaico Templates (new)' from 'Mosaico Template List'
3. Added 5 actions Preview/Edit/Settings/Copy/Delete


The remaining items is - the action should trigger the ang events [here](https://github.com/veda-consulting-company/uk.co.vedaconsulting.mosaico/blob/2.x/ang/crmMosaico.crmstar/TemplateItem.html#L7L12)

ang/afsearchMosaicoTemplateList.aff.html
```diff
@@ -2,7 +2,7 @@
   <div class="af-container af-layout-cols">
     <af-field name="title" />
     <af-field name="base" />
-    <af-field name="category_id" defn="{input_attrs: {multiple: true}}" />
+    <af-field name="category_id" defn="{input_attrs: {multiple: true}}" /><a href='/civicrm/admin/options/mailing_template_category' title="{ts}Click here to configure the categories"><i class="crm-i fa-wrench" aria-hidden="true"></i></a>
```
colemanw: @monishdeb that smarty tag isn't going to work in this tpl, need the angular syntax.
And it needs to have `target="crm-popup"`

ang/crmMosaico/CreateFromBaseTemplate.html
```diff
@@ -0,0 +1,28 @@
+<div id="bootstrap-theme" ng-controller="CrmMosaicoCreateFromBaseTemplate" crm-dialog="crmMosaicoCreatePopup">
+  <form name="crmMosaicoCreateForm">
+
+    <label for="new-tpl-title">{{:: ts('Title') }} <span class="crm-marker">*</span></label>
+    <input required class="form-control" id="new-tpl-title" ng-model="$ctrl.model.title">
+
+    <label for="new-tpl-category">{{:: ts('Category') }}</label>
+    <input class="form-control" id="new-tpl-category" ng-model="$ctrl.model.category" crm-entityref="{entity: 'OptionValue', api: {params: {option_group_id: 'mailing_template_category', is_active: 1}}, select: {allowClear:true, minimumInputLength: 0, placeholder: ts('Select')}}">
```
seamuslee001: @colemanw @monishdeb I was having an issue with the category sticking and had to change this to be $ctrl.model.category_id here instead of category

colemanw: Good catch @seamuslee001. Fixed.

CRM/Mosaico/Upgrader.php
```diff
@@ -120,10 +120,27 @@ public function upgrade_4704() {
...
       ON DELETE SET NULL
     ');
+    
+    CRM_Core_Invoke::rebuildMenuAndCaches(TRUE);
+
+    return TRUE;
+  }
+  
+  /**
+   * Add menu for traditional mailing.
```
lcdservices: not a big deal, but the comment should be updated to reflect the change

colemanw: Fixed.

lcdservices:
This is working as expected. I think it's ready to be merged.

mattwire:
Merging based on feedback above. @colemanw @monishdeb if there is further work/fixes it can be done via follow up PRs before next release

PR #481, opened 2022-02-03
Cleanup BlockPreview.html

colemanw:
Stop relying on deprecated variable CRM.crmMailing.testGroupNames, use static binding with ts() to improve performance.

Once this is merged and released, CiviCRM core can remove the deprecated variable, per https://github.com/civicrm/civicrm-core/commit/d1aefa39fa45d60ccd4f2e0ecdb382ae796c25ef

PR #482, opened 2022-02-08
Cleanup BlockPreview.html (again)

colemanw:
Same as #481 but for the evil twin of the same file.

- Use one-time binding for static strings in ts()
- Add type="button" to button elements
- Make sure labels and form elements match

Same damn cleanup, different damn file. In case I haven't told enough people, I really hate the fact that there are two of every template file in this extension, and I can't wait to get rid of the duplicates.


colemanw:
Thanks @mattwire

PR #484, opened 2022-02-20
Increment version

eileenmcnaughton:
@mattwire I think we should drop this to get @colemanw's change out there - my understanding is it still works with 5.46 but the new version is needed with 5.47 so the sooner we start getting it out there the better

### DIFF
--- a/info.xml
+++ b/info.xml
@@ -15,8 +15,8 @@
     <url desc="Support">https://vedaconsulting.co.uk</url>
     <url desc="Licensing">http://www.gnu.org/licenses/agpl-3.0.html</url>
   </urls>
-  <releaseDate>2021-11-17</releaseDate>
-  <version>2.8</version>
+  <releaseDate>2022-02-21</releaseDate>
+  <version>2.9</version>
   <develStage>stable</develStage>
   <compatibility>
     <ver>5.46</ver>


mattwire:
Ok, we'll need some positive feedback/testing on https://download.civicrm.org/extension/uk.co.vedaconsulting.mosaico/latest/uk.co.vedaconsulting.mosaico-latest.zip before releasing to make sure that the new release doesn't have any obvious problems.

colemanw:
I know that @lcdservices has been using it in production.

monishdeb:
@mattwire I did a clean installation using https://download.civicrm.org/extension/uk.co.vedaconsulting.mosaico/latest/uk.co.vedaconsulting.mosaico-latest.zip on my local and encountered a fatal error on visiting template listing page www.example.com/civicrm/mosaico-template-list :
```
Civi\API\Exception\NotImplementedException: API (MosaicoTemplate, getFields) does not exist (join the API team and implement it!) in Civi\API\Request::create()
```
Found that the Civi/ dir is empty, [MosaicoTemplate](https://github.com/veda-consulting-company/uk.co.vedaconsulting.mosaico/blob/2.x/Civi/Api4/MosaicoTemplate.php)APIv4 entity file is missing.

mattwire:
@colemanw @totten This looks like a packaging issue? Can one of you check? - details here https://docs.civicrm.org/mosaico/en/latest/develop/#publication

colemanw:
@mattwire this should fix it: #487

mattwire:
@monishdeb Can you confirm it is working on a clean install now?

monishdeb:
Yep I tested with https://download.civicrm.org/extension/uk.co.vedaconsulting.mosaico/latest/uk.co.vedaconsulting.mosaico-latest.zip again and it has the Civi and managed files in the ext directory. Then I did the installation on a clean Civicrm setup, all looks good expect a bug (not a blocker though). Let me first share the screencast : 
before

Bug : When you add a new category from message-template page then the new category doesn't appear in the Category dropdown filter.

colemanw:
I agree that's not a blocker, just refresh the page after creating a new category and it will appear.
@mattwire it would be really good to put this release out today so we can start reviewing and merging all the PRs in the queue.

monishdeb:
@colemanw yeah I tried to refresh the page but the new option doesn't appear in the filter, although it shows in the quick edit dropdown: 
before

mattwire:
See #492

colemanw:
@monishdeb that sounds like a caching problem but I couldn't reproduce it. I disabled asset caching and still got the new option when I refreshed the page.

PR #485, opened 2022-02-20
Remove crmStar styling, require Bootstrap3

colemanw:
This removes the complexity of 2 markup styles in favor of Bootstrap3, which is now in CiviCRM core.

Maintaining 2 very-similar-but-slightly-different sets of css styles and html templates was a growing pain. This is a leap forward for maintainability!


colemanw:
@mattwire I'm marking this as a draft for now, but it can be merged as soon as you cut the next release.

colemanw:
@monishdeb this is ready for review.

mattwire:
We'll need to make sure there is plenty of testing before the next release!

PR #489, opened 2022-02-23
Add validation rule to Mosaico uploader

monishdeb:
I tried do it a generic way, like fetch a mutual list of allowed file types from civi's safe_file_extensions option group and mosaico internal library but there seems to be few limitations on doing so. So I went with harcoded file extension which are allowed image ext in both mosaico and civicrm


monishdeb:
@colemanw @lcdservices can you please review this patch !!

PR #490, opened 2022-02-23
Mosaico: loosen restrictions on required fields

monishdeb:
Currently, when building a mailing, you must complete all required fields on the first step of the wizard before you can advance to the builder step. It's not uncommon for one person to be tasked with building out the content for an email, and someone else be responsible for building the recipient list. But those first-step requirements create a problem if the first person to create the email is the content builder. With the traditional builder, you can work on the content as soon as you create the email, and can save it without filling in other required fields. The required fields are only necessary in order to advance to the scheduling step. This patch is to make Mosaico work in similar way

### DIFF
--- a/ang/crmMosaico/BlockMailing.html
+++ b/ang/crmMosaico/BlockMailing.html
@@ -1,9 +1,9 @@
 <div ng-form="subform" crm-ui-id-scope>
 
   <div class="form-group-lg">
-    <label for="inputTitle" class="control-label required-mark">{{ts('Mailing Name')}}
+    <label for="inputTitle" ng-class="checkPerm('create mailings') ? 'control-label' : 'control-label required-mark'">{{ts('Mailing Name')}}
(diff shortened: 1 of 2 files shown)


monishdeb:
needs core PR https://github.com/civicrm/civicrm-core/pull/22818

monishdeb:
@colemanw @lcdservices can you please review this patch !!

colemanw:
Ok, I've merged the core PR, and I checked that the permission is being added clientside (`CRM.checkPerm` doesn't automatically have access to every permission, you have to add them from the `.ang.php` file) here:

https://github.com/civicrm/civicrm-core/blob/6960dd4c1e5e99f8e0b3564dcc15269184d7702d/ang/crmMailing.ang.php#L15-L24

mattwire:
@monishdeb Needs rebase

colemanw:
@mattwire good to merge

monishdeb:
Thank you @colemanw and @mattwire

Issue #483, opened 2022-02-15
Mosaico always interjects the search action for send email (traditional) even if I do not want it


ErikHommel:
I am using a searchTask hook in my extension to remove some tasks. The Mosaico extension however always interjects the "Email - schedule/send via CiviMail (traditional)" task no matter what I have specified. This is because the mosaico extension is at the bottom of the extension list. It should only interject if the standard Email task was in the task list!

This fixes it in mosaico.php, hook searchTasks (but not sure if you want to do it that way):

```
  if ($objectName == 'contact') {
    foreach ($tasks as $task) {
      if ($task['class'] == "CRM_Mailing_Form_Task_AdhocMailing") {
        $tasks[] = [
          'title' => E::ts('Email - schedule/send via CiviMail (traditional)'),
          'class' => 'CRM_Mosaico_Form_Task_AdhocMailingTraditional',
        ];
      }
    }
  }

```

totten:
@ErikHommel Looks like #496 was merged to address this. It seems like a good improvement, though it's a slightly different formulation than the above. Is this closable now?

> using a searchTask hook in my extension to remove some tasks

I don't know the exact context, but just want to mention... if the goal is to do *general filtering* on the task list, you may find that [Symfony Hooks](https://docs.civicrm.org/dev/en/latest/hooks/usage/symfony/) with [Priorities](https://docs.civicrm.org/dev/en/latest/hooks/usage/symfony/#priorities) make the technique more robust.

```php
Civi::dispatcher()->addListener('hook_civicrm_searchTasks', '_my_task_filter', -1000);
function _my_task_filter($event) {
  // Examine $event->objectName and filter $event->tasks
}
```

I skimmed a dozen extensions in `universe` that implement `hook_civicrm_searchTasks` (default priority). All of them were adding tasks, and only half had guards. If you put the filter later (`$priority`=`-1000`), then it should be able to filter-out any task(s) defined by typical extensions.

ErikHommel:
Thanks @totten

PR #496, opened 2022-03-02
Don't add the "schedule/send via CiviMail" search task, if the user has insufficient permissions

bjendres:
The "schedule/send via CiviMail" search task is currently *always* added, however the subsequent action requires the CiviMailing permissions.

With this PR the search task would only be injected if the ``access CiviMail`` permission is granted for the current user.

See also #483


colemanw:
This looks mergeable to me @mattwire

PR #512, opened 2022-04-05
Define customTemplatesUrl absolute as templatesUrl

francescbassas:


francescbassas:
This PR fix this mosaicomsgtpl issue https://lab.civicrm.org/extensions/mosaicomsgtpl/-/issues/26

mattwire:
@francescbassas We should be using `\Civi::settings()->get()` here instead of the deprecated CRM_Core_BAO_Setting. Can you update?

PR #520, opened 2022-05-03
Update civix templates

totten:
See also: https://civicrm.org/blog/totten/civix-v2205-how-remove-million-lines-extra-code

### DIFF
--- a/mosaico.php
+++ b/mosaico.php
@@ -72,24 +72,6 @@ function mosaico_civicrm_upgrade($op, CRM_Queue_Queue $queue = NULL) {
   return _mosaico_civix_civicrm_upgrade($op, $queue);
 }
 
-/**
- * Implements hook_civicrm_angularModules().
- *
- * Generate a list of Angular modules.
- *
- * Note: This hook only runs in CiviCRM 4.5+. It may
- * use features only available in v4.6+.
- *
- * @link http://wiki.civicrm.org/confluence/display/CRMDOC/hook_civicrm_caseTypes
- */
-function mosaico_civicrm_angularModules(&$angularModules) {
-  $canRead = Civi::service('civi_api_kernel')->runAuthorize(
-    'MosaicoTemplate', 'get', ['version' => 3, 'check_permissions' => 1]);
-  if (!$canRead) {
-    return;
-  }
-}
-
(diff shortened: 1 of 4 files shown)


mattwire:
@totten What is the minimum CiviCRM core version requirement for this?

totten:
@mattwire I've tested the civix updates generally on 5.39, 5.44, 5.45, and master. Since `mosaico`'s current `<compatibility>` points to 5.48, this shouldn't change the minimum.

(ED: And I would expect compat further back, but I haven't tested further back.)

colemanw:
@totten as you noted, Mosaico's current `<compatibility>` is 5.48, therefore I don't think any of the shims in this PR are needed, as they are all [in core as of 5.45](https://github.com/civicrm/civicrm-core/pull/22199).

totten:
Correct, they're not required. But I'm ambivalent. There's no runtime cost (CPU/RAM), though there is a passive-storage cost (disk-space). The _practice of curating those files_ also has a cost -- one has to think about the compatibility-matrix and decide when/whether to commit each.

Comparing the size of the files against the curation cost... I don't think it's really worthwhile to fine-tune the list in specific extensions. I'd go for commit+forget.

Of course... from a systemic POV... things do a add-up. It be better if `civix` filtered based on the `<compatibility>`. That's probably a few hours of work though. (We'd need to add some more metadata to each mixin... eg `@providedBy 5.45`)

colemanw:
> (We'd need to add some more metadata to each mixin... eg `@providedBy 5.45`)

I think the `@since` annotation is standard in PHP; it's something we use in other places like APIv4 entities.

totten:
Updated again with 22.05.2. This removes the backport files.

mosaico.php
```diff
@@ -373,6 +324,7 @@ function mosaico_civicrm_mailingTemplateTypes(&$types) {
  * Implements hook_civicrm_entityTypes().
  */
 function mosaico_civicrm_entityTypes(&$entityTypes) {
+  // _mosaico_civix_civicrm_entityTypes($entityTypes);
```
colemanw, line 327: Unnecessary addition.

colemanw:
I don't have merge rights to this repo but :+1: for merging.

PR #541, opened 2022-10-12
Fix crash when upgrading 2.10 via CLI

MegaphoneJon:
This is almost identical to #523 but it happens on upgrade rather than install.


michaelmcandrew:
Got got by this one - thanks for the fix.

PR #561, opened 2023-01-11
Updating to latest Instagram logo

bmango:
Is it possible to update to the latest Instagram logo? The existing one is no longer current.


bmango:
Thanks!!

PR #568, opened 2023-02-06
On WordPress mosaico/img may have a / after - teach the regex to replace this with static images

mattwire:


mattwire:
@seamuslee001 you were right per https://lab.civicrm.org/dev/wordpress/-/issues/138#note_87503

PR #570, opened 2023-02-07
fill empty option field with nbsp 

MarcMichalsky:
With the [CiviCRM Bootstrap Theme](https://github.com/civicrm/org.civicrm.shoreditch) enabled, the Reply-To field collapses in a way that it does not work.

I added a non-breaking space to the default option so that it does not collapse. 

## Before
Bildschirmfoto vom 2023-02-07 14-29-44

## After
Bildschirmfoto vom 2023-02-07 14-30-39


mattwire:
Hi @MarcMichalsky thankyou for contributing. This issue needs to be fixed in the CiviCRM bootstrap theme.

PR #597, opened 2023-06-22
Specify 'match' in mgd.php files

colemanw:
Adds 'match' param to reduce conflicts when resolving managed entities.


colemanw:
Thanks @mattwire

Issue #510, opened 2022-04-04
Translation of TinyMCE


dev-allinappli:
Is there a way to include translation for this file: `packages/mosaico/dist/vendor/tinymce.min.js` because the menu and entries in menu are only in english

PR #603, opened 2023-07-05
issue#510 Add support for TinyMCE translations

mlutfy:
Adds support for translating the TinyMCE editor that is part of Mosaico:

image

Related:

- #510
- https://github.com/voidlabs/mosaico/issues/310
- [Download TinyMCE translations](https://www.tiny.cloud/get-tiny/language-packages/)
- [TinyMCE locale names](https://www.tiny.cloud/docs-4x/configure/localization/)

Ideally we would bundle the translations. Maybe during the build script?

Note that the code is a bit meh with handling the locale codes, because for some reason some use the long-form `xx_YY` and some use the short form, and then it just won't support some locales such as `fr_CA`, even if the translation is there.

Once the PR is merged, to add the translations, they must be copied to either the short of long form, such as:

- `packages/mosaico/dist/tinymce/langs/fr_FR.js`
- `packages/mosaico/dist/tinymce/langs/fr.js`

.. depending on the locale name listed here: https://www.tiny.cloud/docs-4x/configure/localization/

for some locales, we might need in the future to adapt the `getTinymceLocale` function to support more language variants (for now it only supports `fr_CA`).

### DIFF
--- a/CRM/Mosaico/Page/Editor.php
+++ b/CRM/Mosaico/Page/Editor.php
@@ -89,9 +89,21 @@ class CRM_Mosaico_Page_Editor extends CRM_Core_Page {
     // Adding translation strings if exist
     $locale = CRM_Core_I18n::getLocale();
     $lang = CRM_Core_I18n_PseudoConstant::shortForLong($locale);
-    $translationFile = CRM_Core_Resources::singleton()->getPath(E::LONG_NAME, "packages/mosaico/dist/rs/lang/mosaico-{$lang}.json");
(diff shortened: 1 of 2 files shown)


mlutfy:
cc @allinappliadmin  @UshaMakoa @samuelsov

dev-allinappli:
oh yeahhhh! I had opened this issue: https://github.com/veda-consulting-company/uk.co.vedaconsulting.mosaico/issues/510

dev-allinappli:
thanks !

PR #606, opened 2023-07-30
Resolve #572 by ensuring that the value of the base template option i…

### LINKED ISSUE #572: Error when creating a new template with custom base where name is not the same as title
For a client I created a custom base template for them

The base template array looked like this

```php
  $templates['clientname'] = [
    'name' => 'clientname',
    'title' => 'clientname Mosaico Template',
    'path' => E::url('clientname-template/template-clientname.html'),
    'thumbnail' => E::url('clientname-template/edres/_full.png'),
  ];
```

However this seemed to cause a problem in the creation of a new template from this base. 

The issue stems form the fact that the key 'type' is used as the value here https://github.com/veda-consulting-company/uk.co.vedaconsulting.mosaico/blob/2.x/ang/crmMosaico/CreateFromBaseTemplate.html#L21 and is set to the title here https://github.com/veda-consulting-company/uk.co.vedaconsulting.mosaico/blob/2.x/ang/crmMosaico/Templates.js#L16 of the base template however when filtering the list of templates to get the base information it is expecting the name of the base template https://github.com/veda-consulting-company/uk.co.vedaconsulting.mosaico/blob/2.x/ang/crmMosaico/Templates.js#L26 not the title. 

For an easy fix I just change L16 and L31 for my client to be base.name instead of base.title but I would appreciate thoughts on the best way to do this ping @mattwire @totten

seamuslee001:
…s the name of the base template not the title

ping @johntwyman @mattwire

### DIFF
--- a/ang/crmMosaico/CreateFromBaseTemplate.ctrl.js
+++ b/ang/crmMosaico/CreateFromBaseTemplate.ctrl.js
@@ -8,7 +8,7 @@
 
     crmMosaicoTemplates.whenLoaded().then(function() {
       ctrl.bases = _.sortBy(crmMosaicoTemplates.getBases(), 'type');
-      ctrl.model.base = ctrl.bases[0].type;
+      ctrl.model.base = ctrl.bases[0].baseDetails.name;
     });
 
     this.create = function() {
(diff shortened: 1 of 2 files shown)


johntwyman:
We've run into this problem in v2.10.0. The changes proposed in this request have fixed the bug for us.

I consider this a reasonably severe bug - when active it manages to completely break Mosaico in our installation (at least per-domain as we run multisite). We've already deployed the changes here to our production environment accordingly.

PR #617, opened 2023-10-12
Angular - Switch to settingsFactory

colemanw:
Fixes a pattern that was [deprecated back in 2020](https://github.com/civicrm/civicrm-core/commit/d56e4a996cf9de952db80c3823d572d2889365c0)

The "quirky filter" was preserved in a449c021d043dabe7343bf03daa11af2dd9c57f7 as a hedge but I think we can just delete it.

### DIFF
--- a/CRM/Mosaico/Utils.php
+++ b/CRM/Mosaico/Utils.php
@@ -437,4 +437,17 @@ class CRM_Mosaico_Utils {
     flush();
   }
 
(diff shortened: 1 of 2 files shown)


colemanw:
Thanks @mattwire

colemanw:
@mattwire this fixes something that will be broken soon by core (it was deprecated 3 years ago, about to be completely unsupported) so we should do a release soon.

PR #619, opened 2023-10-31
Avoid calling preg_replace on array content

IProSoft:
In our company we are using some extensions and patches to CRM Core that alter CRM_Core_BAO_MessageTemplate::sendTemplate params, adding an array of additional params like event info.

After our changes \Civi\Core\Event\GenericHookEvent $e->content contains one item that is not string, but array so calling preg_replace_callback on all items removes our data.

This change won't affect any of [uk.co.vedaconsulting.mosaico] functions but will save lot work in our side.

**I will be very grateful if you implement this change**


mattwire:
@IProSoft Please can you add a (1 or 2 line) comment to the code to explain why so that a future developer understands and does not undo the changes.

mattwire:
No response from @IProSoft  - please reopen if you are still interested in this change

colemanw:
@IProSoft this was just waiting on a comment from you.

mattwire:

PR #623, opened 2023-12-04
Error in call to mosaico_wrapMailingApi

demeritcowboy:
I'm not sure about this because I don't know how to reproduce it, but on some sites (drupal 9) it doesn't seem to be able to find this function without the backslash in front. The symptom is a blank new mailing page. But e.g. it seems fine on https://d9-master.demo.civicrm.org. But adding the backslash doesn't seem like it would hurt?

Possibly differences in symfony version? Or there was a suggestion drupal multilingual.


mattwire:
I think this is ok @totten ?

mattwire:
@demeritcowboy Still required?

demeritcowboy:
On a site that's been running the patch I took it out and things seem ok so maybe a core update "fixed" it. It's still running the same mosaico version as before (3.3) but core has since been upgraded. Let's close.

PR #627, opened 2023-12-20
Move navigation menu items declarations to managed entities

mlutfy:
Moves the declarations of the Mosaico menu items to  "managed" entity declarations, so that : 

- They can be edited from the admin interface (manage navigation menu)
- They can be positioned more precisely.

image

The clutter in this menu still makes my head explode, but at least the items are placed more logically, and easier to customize by admins. The rest is mostly an issue in core.


mattwire:
Cool. I hadn't realised that moving them to managedentities would allow the navigation menu configurator to work!

PR #236, opened 2018-01-26
Add template for `crm-*` UI's

totten:
# Summary

Introduces an alternative layout based on the `crm-*` CSS convention. Autodetects the best available layout:

* If you have `shoreditch` installed, then the `auto` layout will pick the `bootstrap-wizard` layout.
* If you do *not* have `shoreditch installed, then:
    * The `auto` layout will pick the `crmstar-single` layout.
    * The system-status check will display a gentle notice advising the UI is optimized for use with Shoreditch.

# Screenshots

The appearance in Shoreditch is unchanged. In vanilla `crm-*`, the UI looks a lot like the normal CiviMail UI. This clearly isn't as pretty, but it should allow to phase-in the modules more gently.

*Managing templates:*

crmstar-tpls

*Creating a new mailing (before selecting any templates):*

crmstar-select

*Creating a new mailing (after selecting a templates):*

crmstar-mydesign

*Preview:*

crmstar-preview

# Technical details

For the most part, the main composition screen simply uses the upstream UI components (core's `crmMailing` module). However, the thumbnail gallery and the preview/test dialog needed particular attention.

The file structure has been updated to clearly differentiate the parts that are (a) general, (b) Bootstrap-specific, and (c) `crm-*`- specific. For example:

* Shared/general Angular code: `ang/crmMosaico/` 
* Bootstrap-specific Angular code: `ang/crmMosaico.bootstrap/`
* Bootstrap-specific (S)CSS:  `css/mosaico-bootstrap.css` and  `sass/mosaico-bootstrap*`
* `crm-*`-specific Angular code `ang/crmMosaico.crmstar/`
* `crm-*`-specific (S)CSS: `css/mosaico-crmstar.css` and `sass/mosaico-crmstar*`

To use auto-detection, the `mosaico_layout` setting must be set to `auto`. This will (once again) be the default.

PR #632, opened 2024-01-19
Move away from gulp

artfulrobot:
The project's sass had not been compiled for 2 years, and in that time gulp has changed a lot, and after spending an hour or two trying to get gulp 4 working with the old gulpfile, I gave up.

This PR simplifies the compilation a lot; we now only depend on one npm package, `sass`, and we run it directly - see the compile-sass-to-css.sh file.

Works for me!

I updated the docs, too.

As part of this work I realised that this extension *depends on* shoreditch, but *only for building its css*. 

PS. I built this on top of https://github.com/veda-consulting-company/uk.co.vedaconsulting.mosaico/pull/631 and there was a commit that didn't quite make it into that PR that's included in this. Other than that minor change, I don't think there were any other sass/css changes made.


mattwire:
Nice, thanks @artfulrobot I wonder if it could be made to depend on https://lab.civicrm.org/extensions/theisland instead of shoreditch?

MegaphoneJon:
There is no dependency on Shoreditch anymore: https://github.com/veda-consulting-company/uk.co.vedaconsulting.mosaico/pull/236

mattwire:
@MegaphoneJon Are you sure? There's no dependency at runtime but seems there still might be for compiling the SASS

PR #634, opened 2024-02-14
Fix install collation

colemanw:
See https://github.com/totten/civix/pull/329


mattwire:
@colemanw See my comments at https://github.com/totten/civix/pull/329 I'm weary about merging this because it forces specific collation/row-format instead of relying on database defaults which may cause problems if someone has changed their defaults.

PR #635, opened 2024-02-22
(Draft) Installer - Auto-generate SQL to match local schema

totten:
Remove static SQL files. Generate dynamic SQL during installation. This addresses collation consistency in the same way as CiviCRM's `CustomField`s -- by imitating the live CiviCRM schema on the local database. (This means that the MySQL defaults will not lead to mismatched collations.)

It's an alternative to #634.

The patch is generated by `civix`. I've marked it as "Draft" because it's a demonstration. The corresponding `civix` updates are still pending. (https://github.com/totten/civix/pull/331)


mattwire:
I'm going to close this. As civix now seems to be working well for upgrading to EntityFrameworkV2 we can easily run again next time we are doing a release.

PR #653, opened 2024-07-31
Refactor out calls to deprecated CRM_Utils_Array::value

colemanw:
Deprecated function === bad


mattwire:
@colemanw I think you mean Deprecated function == bad ?

PR #661, opened 2024-11-29
Remove composer.json replaced packages

sam-griff:
In the composer.json there are currently 2 packages marked as replaced:
- guzzlehttp/guzzle ^6.5 || ^7.0
- guzzlehttp/psr7 ^1.8.5

These aren't actually replaced by this extension and are presumably included here as they can be assumed to be provided elsewhere in the civicrm installation, however this can cause conflicts when other requirements conflict with specific versions (e.g. roave/security-advisories conflicts with guzzlehttp/guzzle <6.5.8|>=7,<7.4.5)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,10 +5,6 @@
     "intervention/image": "^2.4",
     "civicrm/composer-downloads-plugin": "^3.0 || ^4.0"
   },
-  "replace": {
-    "guzzlehttp/guzzle": "^6.5 || ^7.0",
-    "guzzlehttp/psr7": "^1.8.5"
-  },
   "config": {
     "allow-plugins": {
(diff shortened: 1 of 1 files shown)


mattwire:
@sam-griff We need these replace statements so the composer library versions don't conflict with CiviCRM core (ie. we are "replacing" those libraries with the CiviCRM core versions).  As such they should track whatever is in the "require" section of CiviCRM core composer.json so happy for them to be updated if necessary.

francescroycampderros:
For me, the proposed solution also works, when trying to install the 3.6 version with Civi 5.82.1 and Drupal 10.4.4

mattwire:

PR #669, opened 2025-03-14
(POC) Dynamic base template

totten:
This is a just an experiment to see how hard it might be to generate a base-template in PHP or Smarty. In theory, that would allow hooks, translations, etc.

The main problem with the current draft: each Mosaico base-template has a folder called `edres` ("_editor resources_", probably) -- these seem to be files used by the admin UI. In this draft, it resolves the `edres` URL incorrectly (which creates some visual distortions in the editor). I haven't dug into fixing, but it feels discrete+fixable (eg several hours work).

Anyway, I thought it would be interesting to share the experiment.


mattwire:

PR #672, opened 2025-05-30
Convert to EntityFrameworkV2

mattwire:


eileenmcnaughton:
@mattwire I believe you are a merger here - are you still testing this?

tests/phpunit/api/v3/Job/MosaicoMigrateTest.php
```diff
@@ -65,7 +65,7 @@ public function testMigrate(): void  {
   }
 
   protected function createExampleLegacyTemplate(): void {
-    $msgTpl = civicrm_api3('MessageTemplate', 'create', [
+    $msgTpl = civicrm_api3('MosaicoMsgTemplate', 'create', [
```
totten, line 68: JFYI @mattwire, this test started failing in Jenkins.

The 1.x=>2.x migration was a long time ago. (Like... 7 years?!) Maybe we can drop everything related to it?

mattwire: Ah, that would simplify things a lot! Had not realised it wasn't used

seamuslee001: also here we are wanting to create a message template not the MosaicoMsgTemplate because that latter bit is done below in the executeQuery becasue there was no APiv3 for it

PR #629, opened 2024-01-09
Fix WSOD on language prefixes

jofranz:
Fix for #567

## Previous behavior
Opening an existing template leads to a white screen of death

## Current behavior
Template can be opened or edited without problems

## Testing
Has been tested with a customer

## Credits
https://github.com/veda-consulting-company/uk.co.vedaconsulting.mosaico/issues/567#issuecomment-1468626246 was suggested fix by @CsarRamos (would you like to test it?)

## Ref
_systopia reference: 23566_

Issue #567, opened 2023-02-02
Drupal multilingual with path prefix, template url


adixon:
Similar to https://github.com/veda-consulting-company/uk.co.vedaconsulting.mosaico/issues/543

On a Drupal 9 multilingual site using path prefixes, the 'template' value stored in the metadata field of civicrm_mosaico_template is getting messed up when the template is saved, rendering it unuseable (i.e., when trrying to use the template, you get the dreaded mosaico "blank" (all grey) screen.

On a mono-lingual site, I can see that the template used to be stored as a url, unescaped, and it's now stored with forward slashes escaped and the domain removed. That's nice as far as being able to swap domains for a dev copy!

What's happening for multilingual sites with a path prefix (and possibly with the related issues including the one above) is that the template url is getting saved as

`en\/\/sites\/default\/files\/civicrm\/ext\/uk.co.vedaconsulting.mosaico\/packages\/mosaico\/templates\/versafix-1\/template-versafix-1.html`

There are two issues here:
1. the double forward slash after the 'en' [which may not be critical]
2. how that path gets implemented when creating a new mailing based on the template, which is the not-found-url
`https://domain.org/en/en//sites/default/files/civicrm/ext/uk.co.vedaconsulting.mosaico/packages/mosaico/templates/versafix-1/template-versafix-1.html`

This could be resolved in either how the template value gets saved, or how it gets used. Since it's broken in use both for using a template and for editing it, I think the main issue is how it's getting stored.

My quick reading of the changes here: https://github.com/veda-consulting-company/uk.co.vedaconsulting.mosaico/pull/529/files
is that urls are being constructed by combining the civicrm base url and the path, which doesn't always work (i.e. when the cms is fairly sophisticated with the request routing).

AlainBenbassat:
I noticed exactly the same issue on a multilingual site.

adixon:
Not Drupal 9 specific, I get the same symptom on a Drupal 7 multilingual with path prefix as well.

On a parallel note - a version of this has been an issue for a while, I have the following aliases in my apache config that works around some parts of this problem:

Alias /en/sites/default/files/  /var/www/drupal/web/sites/default/files/
Alias /fr/sites/default/files/  /var/www/drupal/web/sites/default/files/

AlainBenbassat:
> Not Drupal 9 specific, I get the same symptom on a Drupal 7 multilingual with path prefix as well.
> 
[...]

PR #674, opened 2025-06-10
Fix for issue #567 - Civi cannot handle Drupal multilanguage path, so cannot find template url

kainuk:
Alternative for https://github.com/veda-consulting-company/uk.co.vedaconsulting.mosaico/pull/629

### DIFF
--- a/api/v3/MosaicoTemplate.php
+++ b/api/v3/MosaicoTemplate.php
@@ -73,12 +73,10 @@ function civicrm_api3_mosaico_template_get($params) {
     $metaData = json_decode($getresult['values'][0]['metadata'], TRUE);
     $baseTemplateURL = $metaData['template'];
 
-    $baseURL = CRM_Utils_System::baseURL();
-
     if (_civicrm_api3_mosaico_template_getDomainFrom($baseTemplateURL)) {
       $urlParts = parse_url($baseTemplateURL);
       $templatePath = $urlParts['path'];
-      $currentURL = CRM_Utils_System::baseURL() . $templatePath;
+      $currentURL =  \Civi::paths()->getVariable('cms.root', 'url') . $templatePath;
     } else {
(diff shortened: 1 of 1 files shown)


kcristiano:
https://chat.civicrm.org/civicrm/pl/8ru7afftr7g1bx1itaqbb947zc

I have encountered an issue with a site newly updated to Mosaico 3.7 and hit an issue on   CiviCRM 6.4 as well as 5.81.3  all on WordPess.   Now if we creatre a new mailing from a user created template (based on versafix in our case)     We get this error when trying to use the template
```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://cvdemo.tadpole.ccwp-content/uploads/civicrm/ext/uk.co.vedaconsulting.mosaico/packages/mosaico/templates/versafix-1/template-versafix-1.html. (Reason: CORS request did not succeed). Status code: (null).
```
We've checked the `civicrm.settings.php` to see if the changing `CIVICRM_UF_BASEURL` to have a `/` but it made no difference.   

The issue is in the stored metadata of the template we have not stored the leading slash before `wp-content/uploads/civicrm/ext/uk.co.vedaconsulting.mosaico/packages/mosaico/templates/versafix-1/template-versafix-1.html`  

Reverting the patch fixes the issue.

Alternatively modifying line 79 to

```
$currentURL =  \Civi::paths()->getVariable('cms.root', 'url') . '/' . $templatePath;
```

Also fixes the issue.   I am unsure how either change will affect this issue.   

Change results in the following on WP

before:
`cv php:eval "echo CRM_Utils_System::baseURL();"`

result:

`https://cvdemo.example.org/`

After this PR:

`cv php:eval "echo \Civi::paths()->getVariable('cms.root', 'url') ;"`

result:

`https://cvdemo.example.org`

PR #675, opened 2025-06-11
Ensure that Flexmailer does not re-trigger the processing in onCompose

seamuslee001:
@mattwire 

This should be fine because I believe the weighting here https://github.com/veda-consulting-company/uk.co.vedaconsulting.mosaico/blob/2.x/CRM/Mosaico/Services.php#L41 compared to https://github.com/civicrm/civicrm-core/blob/master/ext/flexmailer/src/Services.php#L64 means that I think Mosaico runs first so we can disable flexmailer's re-processing at least

### DIFF
--- a/CRM/Mosaico/MosaicoComposer.php
+++ b/CRM/Mosaico/MosaicoComposer.php
@@ -14,6 +14,17 @@ class CRM_Mosaico_MosaicoComposer extends \Civi\FlexMailer\Listener\DefaultCompo
     return $mailing->template_type === 'mosaico';
   }
 
+  public function onCompose(
+    \Civi\FlexMailer\Event\ComposeBatchEvent $e
+  ) {
+    if (!$this->isActive() || !$this->isSupported($e->getMailing())) {
+      return;
+    }
+    parent::onCompose($e);
+    // Disable the processing in the parent onCompose function now that we have processed it already
+    \Civi::service('civi_flexmailer_default_composer')->setActive(FALSE);
+  }
+
   public function createTokenProcessorContext(
     \Civi\FlexMailer\Event\ComposeBatchEvent $e
   ) {


djnatedagreat:
I've verified that this works as intended and it suppresses Flexmail's DefaultComposer from firing. Mails seem to be delivered just fine as well.

totten:
Yes, weighting means that `mosaico_flexmail_composer` precedes `civi_flexmailer_default_composer`. (Confirmed with `cv event /flexmail/`.)

Re-reading `DefaultComposer`... There are three guards which decide whether it tries to render content:

* [`isActive()`](https://github.com/civicrm/civicrm-core/blob/0dee43e0ba9b573ad9c0b42dd8cd1bb5ea2e0561/ext/flexmailer/src/Listener/DefaultComposer.php#L57) -- Which does allow third-parties to disable it.
* [`isSupported(...mailing...)`](https://github.com/civicrm/civicrm-core/blob/0dee43e0ba9b573ad9c0b42dd8cd1bb5ea2e0561/ext/flexmailer/src/Listener/DefaultComposer.php#L57) -- This is the part surprises me. It feels like `DefaultComposer::isSupported()` should return `FALSE` for a Mosaico mailing. (But it does appear to have been that way for a long time.)
* [`$hasContent`](https://github.com/civicrm/civicrm-core/blob/0dee43e0ba9b573ad9c0b42dd8cd1bb5ea2e0561/ext/flexmailer/src/Listener/DefaultComposer.php#L71-L82) -- This is where I think the `DefaultComposer` really should decline to process. If the earlier `MosaicoComposer` has generated the message content, then `DefaultComposer` should simply `return`.

Is there some circumstance where `DefaultComposer` is actually generating content?

Or are you looking at some subtler interaction? (e.g. After checking `isSupported()` but before checking `$hasContent`, it does some intermediate steps like reading the message-templates -- perhaps one of those steps has a side-effect?)

CRM/Mosaico/MosaicoComposer.php
```diff
@@ -14,6 +14,17 @@ public function isSupported(\CRM_Mailing_DAO_Mailing $mailing) {
...
 
+  public function onCompose(
+    \Civi\FlexMailer\Event\ComposeBatchEvent $e
+  ) {
+    if (!$this->isActive() || !$this->isSupported($e->getMailing())) {
+      return;
+    }
+    parent::onCompose($e);
+    // Disable the processing in the parent onCompose function now that we have processed it already
+    \Civi::service('civi_flexmailer_default_composer')->setActive(FALSE);
```
totten, line 25: Yeah, this is the way that an extension (like `mosaico`) can fully disable `DefaultComposer`.

One should distinguish between a _complete deactivation_ and a _limited deactivation_. Complete deactivation is simple. I think this is aiming to be limited (*i.e. only disable for the duration of a Mosaico delivery - don't affect any other mailings*), which makes sense.

In that case, it should restore the old value of `$active`. Alas, that requires more SLOC. Maybe a few ways to do a restore. Example pseudocode:

```php
/**
 * Fully disable the DefaultComposer for the duration of Mosaico processing.
 *
 * (This feels like a temporary work-around for some kind of bug; if that's fixed, then delete this.)
 *
 * @service civi.mosaico.default_composer_override
 */
class DefaultComposerOverride extends AutoService implements EventSubscriberInterface {

  public static function getSubscribedEvents() {
    return ['civi.flexmailer.compose' => [
      ['onStartCompose', 2000],
      ['onEndCompose', -2000],
    ]];
  }

  protected $stack = [];

  public function onStartCompose($e) {
    if (/*mailing type is mosaico*/) {
      $c = Civi::service('civi_flexmailer_default_composer');
      $this->stack[] = $c->getActive();
      $c->setActive(FALSE);
    }
  }

  public function onEndCompose($e) {
    if (/*mailing type is mosaico*/) {
      $c = Civi::service('civi_flexmailer_default_composer');
      $c->setActive(array_pop($this->stack));
    }
  }

}
```

mattwire: @totten because you might have mosaico + non-mosaico mailings going out in the same batch/process I guess?

totten: @mattwire Yep.

Which obviously isn't going to happen often. But when it does happen, you'd have blank/omitted email content without any particular clue. (For the afflicted bug-hunter, it would be quite hard to figure the steps-to-reproduce.)

mattwire:
@seamuslee001 Would you be able to look at / make the changes suggested by @totten?

PR #676, opened 2025-06-12
Fix test failure as there is no MosaicoMsgTemplate API v3 entity

seamuslee001:
ping @mattwire


seamuslee001:
ping @totten as well

totten:
@seamuslee001 Yep, I think you are correct. Also, confirmed it runs.

(*I also think it would be fine to remove the subsystem. But this is easier to read.* :upside_down_face:)

PR #681, opened 2025-07-01
Fix Zip building by changing sql for schema given the EntitySchmea v2…

seamuslee001:
… change

@totten @mattwire


totten:
When I run this locally (`./bin/setup.sh -z`), it fixes the build-error about `sql/` -- and the resulting zip file includes both `schema/` and `packages/`. :+1: 

Matt had made a reference to `docs/` in MM -- not sure if he's implying "_We should add `docs/`_" (or if `docs` is just an interesting example folder to illustrate inclusion/exclusion).

mattwire:
Re. docs, no I just noticed it wasn't there along with the other missing directories. Thanks for the PR @seamuslee001

Issue #664, opened 2025-01-25
Sending a test email to multiple recipients separated by comma creates a duplicate contact if there's a space after the comma


demeritcowboy:
e.g. if you send the test to `person1@org1.org,person2@org2.org`, everything's fine. But if you add a space, `person1@org1.org, person2@org2.org`, then person2 still receives it, but a duplicate contact is created for them.

This doesn't happen with "traditional" mailing in core.

francescbassas:
This error has been affecting us for a long time. Thanks for fixing it. Are there any plans to release a new version of the extension that incorporates it?

PR #683, opened 2025-07-08
Use upstream tidy process to fix the send to a group functionality wh…

seamuslee001:
…ilst preserving fix for #664


composerjk:
Testing a send to group seems to work, now. Thanks.

PR #697, opened 2025-10-23
Replace reference to deprecated exception class

colemanw:
The api-specific exception classes have been deprecated and will soon be removed from core.
They are identical to CRM_Core_Exception.


mattwire:

PR #704, opened 2026-07-31
Remove CSS override of core's mailing-preview dialog

### LINKED ISSUE #703: mosaico-bootstrap.css styles core's mailing-preview dialog and breaks it on resize
Open a mailing in the composer, click **Preview as HTML**, then toggle the dialog fullscreen (or resize it). The preview disappears. It does not come back on further resizing — only on closing and reopening the dialog.

This happens on any site with Mosaico installed, including when previewing a **non-Mosaico** mailing.

A PR with the fix is up: **#704**.

## Screenshots

Same mailing, same action.

<table><tr>
<td width="50%">After toggling fullscreen with the rule present: the dialog titlebar is there but the content area is empty and the page behind shows through</td>
<td width="50%">The same action with the rule deleted: the preview renders and stays visible</td>
</tr><tr>
<td align="center"><em>with the rule (current): preview gone</em></td>
<td align="center"><em>with the rule deleted: preview renders</em></td>
</tr></table>

The preview is small in both shots, and stays small after the fix. That is a different problem, unrelated to this rule: core's `crmUiIframe` hardcodes `height="250px"` and then rewrites that attribute from the dialog's own height, which the dialog in turn derives from the iframe — a loop in which neither can grow. Nothing in this report addresses it, and it is not yours to fix; I mention it only so the "after" shot is not read as the fix having failed. What changes here is the preview going from 0px and invisible to present and stable.

## Why it happens

`css/mosaico-bootstrap.css` (and its source, `sass/mosaico-bootstrap/_wizard.scss`) contains:

```css
div[ng-controller=PreviewMailingDialogCtrl] iframe.resized {
  height: calc(100% - 57px);
  position: absolute;
  z-index: -1;
}
```

`PreviewMailingDialogCtrl` is a CiviCRM **core** controller, from `ext/civi_mail/ang/crmMailing/PreviewMgr/{html,full}.html`. The `.resized` class is added by core's `crmUiIframe` directive (`ang/crmUi.js`) on the first `dialogresize`. So the rule sits dormant until core resizes the dialog, then applies to core's preview: `position: absolute` takes the iframe out of flow, leaving the dialog's content box with nothing in it so it collapses to height 0, and `z-index: -1` puts the iframe behind the dialog.

That is also why it reproduces with a traditional mailing — the selector keys on core's controller, not on mailing type. And the content is never lost, only hidden: the iframe's document holds the same bytes throughout.

Measured on CiviCRM 6.16.2 Standalone with Mosaico 3.12.1773937815, in Chrome:

| | iframe height | dialog content height |
|---|---|---|
| on open | 255 | 260 |
| after fullscreen | **0** | **0** |

With the rule removed and nothing else changed, the same dialog reads 260 / 265 after fullscreen.

## Suggested fix

Delete the rule, from both the CSS and its SASS source.

Nothing Mosaico renders can match it: `PreviewMailingDialogCtrl` appears nowhere else in the extension, and `.resized` is added only by core. So the selector's only possible target is core's own preview dialog, where it breaks the layout.

Verified safe by routing Mosaico's own Test-dialog preview back through core's `crmMailingPreviewMgr` with the rule deleted: it renders and survives both fullscreen and drag-resize, in Chrome and Firefox.

If the rule was originally intended to make full-document Mosaico previews fit that dialog, it is worth noting that on current core the directive already sizes the iframe itself on `dialogresize` — the two are fighting, and the CSS wins in a way that removes the element from layout entirely.

RisingOrange:
Open a mailing in the composer, click **Preview as HTML**, then toggle the dialog fullscreen or resize it. The preview disappears, and does not come back on further resizing — only on closing and reopening the dialog. This happens on any site with Mosaico installed, including when previewing a **non-Mosaico** mailing.

The cause is this rule, in `css/mosaico-bootstrap.css` and its SASS source:

```css
div[ng-controller=PreviewMailingDialogCtrl] iframe.resized {
  height: calc(100% - 57px);
  position: absolute;
  z-index: -1;
}
```

`PreviewMailingDialogCtrl` is a **CiviCRM core** controller, not one of Mosaico's, and `.resized` is added by core's own `crmUiIframe` directive on the first `dialogresize`. So the rule sits dormant until core resizes its preview dialog, and then applies to it: `position: absolute` takes the iframe out of flow, leaving the dialog's content box with nothing in it to give it height, and `z-index: -1` puts the iframe behind the dialog. Measured on open vs. after fullscreen, the iframe goes 255px → 0.

This PR deletes the rule from the built CSS and the SASS source both, so it does not return on the next build.

**Verified safe.** `PreviewMailingDialogCtrl` appears nowhere else in this extension and `.resized` is added only by core, so the selector's only possible target is core's dialog. To confirm nothing depended on it, I routed Mosaico's own Test-dialog preview back through core's `crmMailingPreviewMgr` with the rule deleted: it renders and survives both fullscreen and drag-resize, in Chrome and Firefox.

Tested on CiviCRM 6.16.2 with Mosaico 3.12.1773937815.

Same mailing, same action (toggle the dialog fullscreen):

<table><tr>
<td width="50%">With the rule present: the dialog titlebar is there but the content area is empty</td>
<td width="50%">With the rule deleted: the preview renders and stays visible</td>
</tr><tr>
<td align="center"><em>before: preview gone</em></td>
<td align="center"><em>after: preview renders</em></td>
</tr></table>

The preview is small in both, which is a separate core sizing issue described in #703 and not addressed here.

Fixes #703.

### DIFF
--- a/css/mosaico-bootstrap.css
+++ b/css/mosaico-bootstrap.css
@@ -133,12 +133,6 @@
   margin-top: 30px;
 }
 
-div[ng-controller=PreviewMailingDialogCtrl] iframe.resized {
-  height: calc(100% - 57px);
-  position: absolute;
-  z-index: -1;
-}
-
 #bootstrap-theme .crm-mosaico-selected.center-block {
   float: none;
 }
--- a/sass/mosaico-bootstrap/_wizard.scss
+++ b/sass/mosaico-bootstrap/_wizard.scss
@@ -42,9 +42,3 @@
 .mosaico-templates-wrapper {
   margin-top: 30px;
 }
-
-div[ng-controller="PreviewMailingDialogCtrl"] iframe.resized {
-  height: calc(100% - 57px);
-  position: absolute;
-  z-index: -1;
-}


RisingOrange:
Closing: development moved to lab.civicrm.org, so this would not have been seen here. Re-submitted as https://lab.civicrm.org/extensions/mosaico/-/merge_requests/725 (against `2.x`, same +0/-12 deletion), for https://lab.civicrm.org/extensions/mosaico/-/issues/707.

